PR #65, opened 2022-08-26
feat(app): add AccountItem component

### LINKED ISSUE #19: Create AccountItem component
## Figma Reference
https://www.figma.com/file/uEckQY4EaQ5rXwyISg6d1g/Fuel-Wallet?node-id=290%3A16192

## Props
- `account: Account`
- `isSelected: boolean`
- `isHidden: boolean`

pedronauck:
✨ Fix #19


packages/app/src/types.ts
```diff
@@ -1,3 +1,12 @@
 export enum Pages {
   'home' = '/',
 }
+
+/**
+ * This should be generated automatic with Codegen and GraphQL endpoints
```
luizstacio, line 6: The accounts and all data are going to be saved locally on the browser. I think the idea of having the types is not temp but permanent.

luizstacio (changes requested):

luizstacio (approved):

PR #76, opened 2022-08-29
feat: add Home screen

### LINKED ISSUE #10: Create wallet Home screen
## Figma reference
https://www.figma.com/file/uEckQY4EaQ5rXwyISg6d1g/Fuel-Wallet?node-id=290%3A16364

## Blocked by
- https://github.com/FuelLabs/fuels-wallet/issues/7
- https://github.com/FuelLabs/fuel-ui/issues/120
- https://github.com/FuelLabs/fuels-wallet/issues/6
- https://github.com/FuelLabs/fuels-wallet/issues/9
- https://github.com/FuelLabs/fuels-wallet/issues/8

LuizAsFight:
- Add Home Screen
- Upgrade TS SDK
- Upgrade fuel-core
- Upgrade faucet
- Add API mock to storybook

Closes #10


packages/app/src/systems/Asset/hooks/useAsset.tsx
```diff
@@ -0,0 +1,14 @@
+import { ASSET_LIST } from "../utils";
+
+export function useAsset(assetId: string) {
+  const asset = ASSET_LIST.find((asset) => asset.assetId === assetId);
```
pedronauck, line 4: You can use an `useMemo` here with the `assetId` to avoid unnescessary recomputations

packages/app/src/systems/Home/hooks/useHome.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+import type { HomeMachineState } from "../machines/homeMachine";
+import { homeMachine } from "../machines/homeMachine";
+
+const selectors = {
+  context: (state: HomeMachineState) => state.context,
+};
+
+export function useHome() {
+  const [state, , service] = useMachine(homeMachine);
```
pedronauck, line 11: Idk, but IMHO this shouldn't be an `useHome` hook in order to get accounts and balances. Since the account will be something we will need to use globally makes more sense we have something like `useAccount` and use it around all the pages. By having an `useHome` to fetch account and balances, them in another screens you will need to have another `useSomething` that will make the same things, you know? 

You can move this `useHome` and `homeMachine` to an `useAccount` and `accountMachine`

pedronauck: And another point is I think we should use `react-query` to just make simple async operations in our database, it handles better with cache in this case and will avoid a bunch of LOC's. Maybe let XState for situations when we have medium to complex state cases. Idk, what do you think @luizstacio ?

packages/app/src/systems/Home/pages/Home/Home.tsx
```diff
@@ -1,10 +1,41 @@
...
+
+  const balances = [...(context.account?.balances || [])];
+
   return (
-    <Layout title="Home">
+    <Layout title="Home" isLoading={state.hasTag("loading")}>
       <Layout.TopBar />
-      <Layout.Content>Helo</Layout.Content>
+      <Layout.Content>
+        <Flex direction="column" css={{ height: "100%" }}>
```
pedronauck, line 20: Let's try to make styles just in CSS props

packages/app/src/systems/Home/pages/Home/Home.tsx
```diff
@@ -1,10 +1,41 @@
...
+          {state.hasTag("loadingAccount") || !context.account ? (
```
pedronauck, line 21: Here you can add this in an XState selector to avoid re-renders.

```jsx
const selectors = {
  isLoading: (state) => state.hasTag('loadingAccount') || !state.context.account
}

const useFoo() => {
  const [state, ,service] = useMachine(fooMachine)
  const isLoading = useSelector(service, selectors.isLoading)
  
  // and here return it as context
  return {
    context: {
      isLoading
    }
  }
}
```

packages/app/src/systems/Home/pages/Home/Home.tsx
```diff
@@ -1,10 +1,41 @@
+import { Flex } from "@fuel-ui/react";
+
+import { useHome } from "../../hooks/useHome";
+import { AssetsTitle } from "../components/AssetsTitle/AssetsTitle";
+import { HomeActions } from "../components/HomeActions/HomeActions";
+
+import { BalanceWidget } from "~/systems/Account/components/BalanceWidget/BalanceWidget";
```
pedronauck, line 7: Don't forget to create `index.tsx` in specifics folder to don't have big imports like this, but have something like this instead:
```js
import { BalanceWidget } from '~/systems/Account'
```

Also, the same on `line 4-5` on `AssetsTitle` and `HomeActions` imports above. Should be:
```js
import { AssetsTitle, HomeActions } from '../components'
```

pedronauck (changes requested):

packages/app/src/systems/Home/pages/components/HomeActions/HomeActions.tsx
```diff
@@ -0,0 +1,43 @@
+import { Button, Flex } from "@fuel-ui/react";
+
+export type HomeActionsProps = {
+  isDisabled?: boolean;
+};
+
+export const HomeActions = ({ isDisabled }: HomeActionsProps) => {
```
pedronauck, line 7: Idk, but I think this is kinda an "over abstraction", these buttons will remain in home always and don't have a specific case to be tested and split into a component

pedronauck (changes requested):

pedronauck (changes requested):
Screen Shot 2022-09-05 at 1 50 39 PM

Have an weird space before balance numbers is balance widget

pedronauck:
Screen Shot 2022-09-05 at 1 53 00 PM

I guess is missing some cases on `Home.stories.tsx`, like with assets, etc.

packages/app/package.json
```diff
@@ -33,6 +35,8 @@
     "events": "^3.3.0",
     "fake-indexeddb": "^4.0.0",
     "fuels": "^0.14.0",
+    "msw": "^0.47.1",
```
luizstacio, line 38: This should go on `devDependencies`?

packages/app/src/systems/Account/machines/accountsMachine.tsx
```diff
@@ -17,74 +17,133 @@ type MachineServices = {
...
+    {
+      // eslint-disable-next-line @typescript-eslint/consistent-type-imports
+      tsTypes: {} as import("./accountsMachine.typegen").Typegen0,
+      schema: {
+        context: {} as MachineContext,
+        services: {} as MachineServices,
+        events: {} as MachineEvents,
+      },
+      predictableActionArguments: true,
+      id: "(machine)",
```
luizstacio, line 42: This should include something like `(machine).<name>` or it just `(machine)`?

packages/app/src/systems/Core/components/PublicRoute/PublicRoute.tsx
```diff
@@ -9,20 +8,14 @@ type PublicRouteProps = {
...
 
-  if (isLoading) {
-    return (
-      <BoxCentered minWS minHS>
-        <Spinner />
-      </BoxCentered>
-    );
-  }
-  if (!isSignUp && !isLoading && accounts?.length) {
+  if (!isSignUp && accounts != null && accounts?.length) {
```
luizstacio, line 16: Why remove the loading state? It's in another place now?

packages/app/src/systems/Home/components/HomeActions/HomeActions.tsx
```diff
@@ -0,0 +1,43 @@
...
+    <Flex
+      css={{
+        marginTop: "$8",
+        marginBottom: "$6",
+        flexShrink: 0,
+      }}
+    >
+      <Button
+        isDisabled={isDisabled}
+        css={{
```
luizstacio, line 18: If we move the style to a object outside here, it can be reused on both buttons

packages/app/src/systems/Home/pages/Home/Home.tsx
```diff
@@ -1,10 +1,42 @@
...
+            <AssetList assets={currentAccount?.balances || []} />
+          )}
+          {Boolean(!isLoading && !currentAccount?.balances?.length) && (
+            <Flex css={{ flex: "1 0" }}>
+              <AssetList.Empty />
+            </Flex>
+          )}
+        </Flex>
+      </Layout.Content>
     </Layout>
```
luizstacio, line 40: This code looks to have room for improvement. Maybe Layout could work as a provider and receive isLoading state.
The component `BalanceWidget`, and `AssetList`.... could have an upper component that uses an `unloading form the layout provider` and renders the correct component. ping @pedronauck here.

luizstacio (changes requested):

packages/app/src/systems/Core/components/PublicRoute/PublicRoute.tsx
LuizAsFight: It was triggering re-renders when you refetch account info

Home page is taking care of loading

packages/app/src/systems/Account/machines/accountsMachine.tsx
LuizAsFight: it's only (machine). same way was done in swayswap and in `signUpMachine` as well

luizstacio (approved):

pedronauck (approved):

PR #82, opened 2022-09-01
chore(app): add unit tests on components and machines

pedronauck:
This pull request intends to add unit tests using Jest in all components and machines


packages/app/jest.config.ts
```diff
@@ -1,31 +1,28 @@
+/* eslint-disable import/no-relative-packages */
+/* eslint-disable @typescript-eslint/ban-ts-comment */
+/* eslint-disable import/order */
 import type { Config } from '@jest/types';
-import { resolve } from 'path';
+// @ts-ignore
+import baseConfig from './node_modules/@fuel-ui/test-utils/dist/config';
```
luizstacio, line 6: Why do we need to import like this?

packages/app/src/systems/SignUp/components/CreatePassword/CreatePassword.tsx
```diff
@@ -60,45 +66,70 @@ export function CreatePassword({
...
+            control={control}
+            render={({ field, fieldState }) => (
+              <Form.Control isInvalid={Boolean(fieldState.error)}>
+                <Form.Label htmlFor="password">Password</Form.Label>
+                <InputPassword
+                  {...field}
+                  id="password"
+                  placeholder="Type your password"
+                />
+                {fieldState.error?.message && (
```
luizstacio, line 80: This will be a common thing. Can we split this logic into a component?

luizstacio (changes requested):

packages/app/jest.config.ts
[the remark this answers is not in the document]
pedronauck: I was with a problem to import this, but since the new `fuel-ui` version now it's working, changed ✅

pedronauck: ✅

.github/workflows/pr.yaml
```diff
@@ -34,3 +34,22 @@ jobs:
       - run: |
           pnpm ts:check
           pnpm lint
+          pnpm test
```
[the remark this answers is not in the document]
pedronauck: ✅

packages/app/src/systems/SignUp/components/CreatePassword/CreatePassword.tsx
pedronauck: ✅

PR #86, opened 2022-09-02
feat: add e2e tests

### LINKED ISSUE #85: Add Cypress and first E2E tests

pedronauck:
✨ Fix #85


.github/actions/setup-rust/action.yaml
```diff
@@ -0,0 +1,36 @@
+name: 'Setup Rust env'
```
luizstacio, line 1: I think we can remove this part as we don't build contracts on the wallet project

packages/app/cypress/support/e2e.ts
```diff
@@ -0,0 +1,20 @@
...
+// 'supportFile' configuration option.
+//
+// You can read more here:
+// https://on.cypress.io/configuration
+// ***********************************************************
+
+// Import commands.js using ES2015 syntax:
+import './commands';
+
+// Alternatively you can use CommonJS syntax:
```
luizstacio, line 19: I think we can remove as we just need ES2015

luizstacio (changes requested):

packages/app/cypress/support/e2e.ts
pedronauck: ✅

.github/actions/setup-rust/action.yaml
pedronauck: ✅

luizstacio (approved):

PR #97, opened 2022-09-10
chore(app): implement mediator/subscribe to avoid global states

pedronauck:
This pull request intends to implement a mediator/subscribe pattern in order to avoid global states and make it possible for us to communicate between our systems without creating a coupled system.

Alongside this implementation we get several benefits:

- Since the [service isn't started internally in React](https://github.com/FuelLabs/fuels-wallet/blob/refac/mediator/packages/app/src/systems/Core/utils/mediator/useMediator.tsx#L8-L14), you don’t need a provider, that’s a huge point!
```jsx
export function App() {
  return (
    <>
      <Counter />
      <Increment />
      <Decrement />
    </>
  )
}
```

- You can subscribe for an event in any component, and take action based on this event using a callback
- Systems don’t talk to each other, they talk with a mediator machine, and the mediator does the work. So, we don’t have coupled system in this way.

```jsx
import { subscribe } from '@fuels-wallet/mediator'

subscribe(events.increment, (num) => {
  setCount((count) => count + num)
})
```

- All events are created as functions and can be in the scope of their own system, so if a system is deleted, is easy to map when we need to delete that event since Typescript will complain that some event doesn’t exist anymore

```tsx
// events.ts
import { createEvent } from "@fuels-wallet"

/**
 * Add a `{system}:` name before the event is a good way to don't have conflicts
 */
enum Events {
  increment = "counter:INCREMENT",
  decrement = "counter:DECREMENT",
}

export const increment = createEvent<number>(Events.increment)
export const decrement = createEvent<number>(Events.decrement)
```

- You can send an event from anywhere in code, you don’t need a hook or an internal react thing to dispatch to the mediator
- All events are well-typed because they are functions created using a factory

```jsx
// component.ts
import * as events from "./events"

export function Increment() {
  return <button onClick={() => events.increment(2)}>Increment</button>
}
```

I really believe this is the better way to don't need a global provider that triggers a lot of re-renders and re-updates. This improves performance and our architecture since we can have scoped/more local states that don't know another part of the software.


packages/app/src/systems/Account/index.tsx
```diff
@@ -1,3 +1,6 @@
+import * as accountEvents from "./events";
+
 export * from "./components";
 export * from "./services";
 export * from "./types";
+export { accountEvents };
```
luizstacio, line 6: `export * as accountEvents from "./events"`?

luizstacio (changes requested):

packages/app/src/systems/Account/index.tsx
pedronauck: ✅

luizstacio (approved):

PR #108, opened 2022-09-15
feat: Faucet flow

### LINKED ISSUE #27: Create Faucet popup component
## Observation
- Use the same implementation [we did on SwapSwap](https://github.com/FuelLabs/swayswap/tree/master/packages/app/src/systems/Faucet)

### LINKED ISSUE #103: Total balance should show total ETH amount
In this case, bellow the user balance should be 0.5 ETH
Screen Shot 2022-09-14 at 3 22 14 PM

LuizAsFight:
- add faucet flow
- fix home balance
- add mock to jest

Closes #27 
Closes #103


packages/app/src/systems/Home/routes.tsx
```diff
@@ -10,7 +12,10 @@ export const homeRoutes = (
     element={
       <PrivateRoute>
         <Home />
+        <Outlet />
       </PrivateRoute>
     }
-  />
+  >
+    <Route path={Pages.faucet} element={<FaucetDialog />} />
```
pedronauck, line 19: The modal here is with almost the entire width of the screen, which should be good to decrease it according to the Figma layout :)

CleanShot 2022-09-15 at 23 05 55@2x

pedronauck (changes requested):
I guess you could also add the Faucet flow inside Cypress!

packages/app/package.json
```diff
@@ -30,6 +30,7 @@
     "@hookform/resolvers": "^2.9.8",
     "@xstate/react": "^3.0.1",
     "buffer": "^6.0.3",
+    "cross-fetch": "^3.1.5",
```
luizstacio, line 33: Why do we need cross-fetch?

packages/app/src/systems/Asset/components/AssetList/AssetList.tsx
```diff
@@ -21,7 +21,8 @@ export const AssetList: AssetListComponent = ({ assets, isLoading }) => {
   if (isLoading) return <AssetList.Loading items={4} />;
 
   const isEmpty = !assets || !assets.length;
-  if (isEmpty) return <AssetList.Empty />;
+  // TODO: how we're gonna identify if it's devnet?
```
luizstacio, line 24: We are going to create a JSON file with all the possible networks

luizstacio (changes requested):

luizstacio:
1. We should create a container to avoid layout re-sizing between loading and showing the captcha;
https://user-images.githubusercontent.com/3941923/191124425-9a13b42e-3436-4c09-bc50-2d5a2d72a7f8.mov
2. The feedback message could be a toast, not on the button, closing the screen right after the request is made.

packages/app/package.json
LuizAsFight: I got the idea from `swayswap`, makes fetch platform agnostic allowing to use in node envs, like testing in jest

packages/app/src/systems/Home/routes.tsx
LuizAsFight: skipping this because it's not possible to resize captcha iframe and get the size we want

luizstacio (approved):

Issue #59, opened 2022-08-24
Create a PoC for the wallet as browser extenssion


luizstacio:
Embed the Wallet application in a browser extension.

PR #120, opened 2022-09-23
feat(app): add network flow

pedronauck:
✨ Fix #44 
✨ Fix #17
✨ Fix #18
✨ Fix #16
✨ Fix #15

PR #121, opened 2022-09-24
feat: create wallet extension

luizstacio:
## Description

This PR creates a basic structure for the CRX Wallet, with;
- When the user installs the extension, open the sign-up page
- If the user tries to open the wallet, open the sign-up page
- Centralize routes logic on a single file `src/routes.tsx` this will help the future split CRX and WebApp into different packages. With modular private/public checking, it makes it impossible for other applications to decide on the private/public route system.

## Installing the extension
1. Access the latest preview build: Ex.: https://fuels-wallet-ed2brkria-fuel-labs.vercel.app/
2. Download the wallet (zip file)
3. Open chrome extension management
4. Enable developer mode (switch on the top right)
5. Drag and drop the zip file
6. Extension installed

close: #59 

## Issues derived from this work
- https://github.com/FuelLabs/fuels-wallet/issues/123
- https://github.com/FuelLabs/fuels-wallet/issues/124
- https://github.com/FuelLabs/fuels-wallet/issues/125


luizstacio:
Holding before #120 is merged

packages/app/manifest.config.ts
```diff
@@ -0,0 +1,36 @@
+import { defineManifest } from '@crxjs/vite-plugin';
+
+export default defineManifest({
+  manifest_version: 3,
+  name: 'Fuel Wallet',
+  version: '1.0.0',
```
pedronauck, line 6: could this be maybe `v0.1.0` since we're in an alpha version yet?

packages/app/src/main.tsx
```diff
@@ -1,11 +1,22 @@
 import '@fontsource/source-code-pro';
+import type { ReactNode } from 'react';
 import { createRoot } from 'react-dom/client';
-import { BrowserRouter } from 'react-router-dom';
+import { BrowserRouter, HashRouter } from 'react-router-dom';
 
 import { App } from './App';
+import { IS_CRX } from './config';
+
+const RouterProvider = ({ children }: { children: ReactNode }) => {
```
pedronauck, line 9: We have another component that needs to make some routing logic that's `TestWrapper`, now with this another routing logic here, we can start to have some confusion around this. Maybe we could have a `Core/Router` component that handles all this inside of it and could be used in all parts of the application: as a wrapper for tests on jest, as a router for CRX and the web.

packages/app/src/systems/Core/components/RouteGuard/RouteGuard.tsx
```diff
@@ -0,0 +1,28 @@
...
+  reject: ReactNode;
+  cond?: (isLogged: boolean) => boolean;
+};
+
+export const guards = {
+  isLoggedIn: (i: boolean) => i,
+  isNotLoggedIn: (i: boolean) => !i,
+};
+
+export function RouteGuard({
```
pedronauck, line 16: good job here 👏🏻

pedronauck (changes requested):

packages/app/src/main.tsx
luizstacio: I was looking at the TestWrapper is used directly on the tests, don't see how to combine it, or if was the idea. But I moved the router provider to inside the system/core

pedronauck (approved):

PR #134, opened 2022-10-05
feat(app): add TxFromTo component

### LINKED ISSUE #34: Create TxFromTo component
## Figma Reference
https://www.figma.com/file/uEckQY4EaQ5rXwyISg6d1g/Fuel-Wallet?node-id=93%3A8300

## Props
- `state: 'default' | 'success' | 'pending' | 'failed'`
- `fromAddress: String`
- `toAddress: String`

## Subcomponents
- `<TxFromTo.Loading>`
  - This component is just a loading state using our Skeleton

## Blocked by
- #31

pedronauck:
✨ Fix #34

### DIFF
--- a/packages/app/src/systems/Transaction/components/TxRecipientCard/TxRecipientCard.tsx
+++ b/packages/app/src/systems/Transaction/components/TxRecipientCard/TxRecipientCard.tsx
@@ -1,15 +1,15 @@
 import { cssObj } from '@fuel-ui/css';
 import { Avatar, Box, Card, Copyable, Flex, Icon, Text } from '@fuel-ui/react';
+import { isBech32 } from 'fuels';
 import type { FC } from 'react';
 
 import { TxRecipientCardLoader } from './TxRecipientCardLoader';
 
-import type { Account } from '~/systems/Account';
 import { shortAddress } from '~/systems/Core';
 
 export type TxRecipientCardProps = {
-  account?: Account;
-  contract?: { address: string };
+  address: string;
+  isReceiver?: boolean;
 };
 
 type TxRecipientCardComponent = FC<TxRecipientCardProps> & {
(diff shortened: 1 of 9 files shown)


LFSCamargo (approved):

packages/app/src/systems/Transaction/components/TxRecipientCard/TxRecipientCard.tsx
```diff
@@ -1,48 +1,48 @@
...
 };
 
 export const TxRecipientCard: TxRecipientCardComponent = ({
-  account,
-  contract,
+  address,
+  isReceiver,
 }) => {
-  const address = account ? account.address : contract?.address;
+  const isAccount = isBech32(address);
```
luizstacio, line 23: I think this will not work, as Tx's will always return a 0x000... key.

luizstacio: The best ways I imagine to know if and address is a contract is;
1. Check on the tx inputs if the address is inside a contract input type.
2. Make a query to the provider. If it returns, the id is a contract if not is a wallet.
```
query {
  contract(id: "0xb4969141d7418a8c84c41a1abd3322fbe0ef9804a4bc35f201427d6b1256ac4a") {
    id
  }
}
```

luizstacio (changes requested):

packages/app/src/systems/Transaction/components/TxRecipientCard/TxRecipientCard.tsx
pedronauck: Ok, so I'll let this job for the transaction screen itself in order to avoid a lot of requests in small components, and to be able to handle this I'll change the way both `TxRecipientCard` and `TxFromTo` receives props, right?!

luizstacio: Yes. Enabling to receive a type on the component will make the work because the Transaction Object will have this info.

luizstacio (approved):

PR #140, opened 2022-10-11
feat(app): add Unlock dialog

### LINKED ISSUE #139: Create a lock and unlock flow
When signing transactions or messages, you need to provide the password to unlock the WalletManager. We should create a screen with the same looking feel as the password creation. This screen should;
- Appear before any action that requires a privateKey;
- After x time the WalletManager should be locked again.

pedronauck:
✨ Fix #139

## Notes

I spent some time trying to add E2E tests for this screen, but I had some problems
- There isn't a good way to mock or handle IndexedDB inside Cypress, we need to add the created vault inside the IndexedDB, then retrieve it using the `AccountService` to unlock the wallet and any of this is possible by using Cypress
- I gave it a try to Playwright as well, but it was getting `import.meta` error and I didn't find anything related on how to fix that or maybe change its bundler configuration.
- Also with Jest, I'm getting the same error I was getting in the past because of `fuels-ts`, something related to this:
CleanShot 2022-10-10 at 20 27 09@2x

So, I ended up with no options for this E2E test 😅

### DIFF
--- a/packages/app/src/config.ts
+++ b/packages/app/src/config.ts
@@ -18,6 +18,7 @@ export const WALLET_HEIGHT = 600;
 export const TAB_BAR_HEIGHT = 30;
 export const IS_CRX = VITE_CRX === 'true';
 export const IS_LOGGED_KEY = 'fuel__isLogged';
+export const IS_LOCKED_KEY = 'fuel__isLocked';
 export const IS_DEVELOPMENT = process.env.NODE_ENV !== 'production';
 export const IS_CRX_POPUP =
   IS_CRX && globalThis.location.pathname === CRXPages.popup;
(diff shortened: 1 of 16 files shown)


luizstacio:
1. Should we disable the navbar? Make something fullscreen?
Screen Shot 2022-10-11 at 12 54 06 PM
2. For E2E tests, have you tried to use playwright instead? I think it enables access indexedb directly

LuizAsFight:
> 1. Should we disable the navbar? Make something fullscreen?
> 
> Screen Shot 2022-10-11 at 12 54 06 PM

I think we should hide "Lock" and "menu" icon from topbar whenever it's an internal page

LuizAsFight (commented):
code looks good. only layout comment

LuizAsFight (approved):

PR #146, opened 2022-10-18
feat: sign message

### LINKED ISSUE #45: Create Signature Request screen
## Figma Reference
https://www.figma.com/file/uEckQY4EaQ5rXwyISg6d1g/Fuel-Wallet?node-id=290%3A15706

## Blocked by
- #44 
- #43

LuizAsFight:
- sign message flow
- create `unlockMachine` and use it as a child in `signMachine`

Closes #45 

`txApproval` is gonna be broken until  `unlockMachine` code changes are applied(Ref #158 and #159), which is gonna be my immediate next PR.


packages/app/src/systems/DApp/hooks/useSignatureRequest.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+const selectors = {
+  isUnlocking: (state: SignMachineState) => state.matches('unlocking'),
+  signedMessage: (state: SignMachineState) => state.context.signedMessage,
+  isUnlockingLoading: (state: UnlockMachineState) => state.matches('unlocking'),
+};
+
+export function useSignatureRequest() {
+  const { account } = useAccount();
+  const [state, send, service] = useMachine(signMachine);
```
luizstacio, line 16: Why `useMachine` instead of `useInterpreter`?

packages/app/src/systems/DApp/pages/TxApprove/TxApprove.tsx
```diff
@@ -32,7 +32,7 @@ export function TxApprove({ id, url }: TxApproveProps) {
     <Layout title="New Transaction" isLoading={isLoading}>
       <Layout.TopBar type={TopBarType.external} />
       <Layout.Content css={styles.content}>
-        <UnlockDialog isOpen={Boolean(isLocked)} isFullscreen />
+        <UnlockDialog isFullscreen onUnlock={() => {}} isOpen={false} />
```
luizstacio, line 35: Maybe remove completely the `onUnlock` if is not needed

packages/app/src/systems/Transaction/hooks/useTransaction.tsx
```diff
@@ -23,7 +23,9 @@ const selectors = {
 };
 
 export function useTransaction(id: string) {
-  const { isLocked, isLoading, wallet } = useAccount();
+  const wallet = undefined;
+  const isLocked = true;
```
luizstacio, line 27: Why have these two been set as mocked values here?

luizstacio (changes requested):

packages/app/src/systems/DApp/hooks/useSignatureRequest.tsx
LuizAsFight: for local machines we normally use `useMachine`
it's good because it returns `[state, send, service]`
`useInterpret` returns only `service`


we need access to child `state.children.unlock` machine here

we used before `useInterpret` for global machines only, but was replaced by new `store` implementation


although `useInterpret` seems better for performance:
> The useInterpret is useful when you want fine-grained control, e.g. minimize re-renders. In contrast to useMachine that would flush each update from the machine to the React component, useInterpret instead returns a static reference (to just the interpreted machine) which will not rerender when its state changes.


if we find other way of accessing child machines, could try moving to `useInterpret`

packages/app/src/systems/DApp/pages/TxApprove/TxApprove.tsx
LuizAsFight: onUnlock is needed, but will be implemented in `txApprove` in next PR:

> `txApproval is gonna be broken until unlockMachine code changes are applied(Ref https://github.com/FuelLabs/fuels-wallet/issues/158 and https://github.com/FuelLabs/fuels-wallet/issues/159), which is gonna be my immediate next PR.`

packages/app/src/systems/Transaction/hooks/useTransaction.tsx
LuizAsFight: we don't have global `isLocked` and `wallet` anymore.
will be implemented in `txApproval` like it is in `signMessage`.

> txApproval is gonna be broken until unlockMachine code changes are applied(Ref https://github.com/FuelLabs/fuels-wallet/issues/158 and https://github.com/FuelLabs/fuels-wallet/issues/159), which is gonna be my immediate next PR.

packages/app/src/systems/DApp/hooks/useSignatureRequest.tsx
luizstacio: Does the state returned from the useSelector(service) doesn't have access to the children's state?

LuizAsFight: it has access but doesn't update as expected.

needs to pass directly the child machine to `useSelector` in order to update

luizstacio (approved):

LFSCamargo (approved):

PR #160, opened 2022-10-21
chore: use process.env instead of import.meta

pedronauck:
This pull request intends to change `import.meta` and use `process.env` instead, mainly because Playwright don't play well with `import.meta` and in the future, we will use Playwright instead of Cypress.


luizstacio:
What is the mean issue with meta.env on playwright to use inside the test files?

LFSCamargo (dismissed):

LFSCamargo (commented):
Just Check Playwright Tests!

luizstacio:
#168

Issue #38, opened 2022-08-24
Create Transaction screen


luizstacio:
## Figma reference
https://www.figma.com/file/uEckQY4EaQ5rXwyISg6d1g/Fuel-Wallet?node-id=290%3A15917

## URL Params
- `transactionId: String`

## Blocked by
- https://github.com/FuelLabs/fuels-wallet/issues/31
- https://github.com/FuelLabs/fuels-wallet/issues/32
- https://github.com/FuelLabs/fuels-wallet/issues/34
- https://github.com/FuelLabs/fuels-wallet/issues/35
- https://github.com/FuelLabs/fuels-wallet/issues/36
- https://github.com/FuelLabs/fuels-wallet/issues/37
- #303 
- #291 
- #290 
- #289 
- #288 
- #286 
- #284

PR #165, opened 2022-10-22
feat: Approve Transaction

### LINKED ISSUE #158: refact `TxApprove`  to use new `unlockMachine`
### Motivation

We should include new `unlockMachine` in tx approval flow

### Usage example

_No response_

### Possible implementations

_No response_

### LINKED ISSUE #177: Approve Transaction flow
### Motivation

We had some changes in approve transaction flow, should apply design changes, this is the mockup:
https://www.figma.com/file/uEckQY4EaQ5rXwyISg6d1g/Fuel-Wallet?node-id=1446%3A20366

Also need to:
- pre-run the transaction to get gas estimations
- add not enought balance error
- add general error feedback
- add tx-approval error screen
- include block-explorer link to Success feedback

### Usage example

_No response_

### Possible implementations

_No response_

LuizAsFight:
Closes #158 
Closes #177


packages/app/src/systems/Asset/components/AssetsAmount/AssetsAmount.test.tsx
```diff
@@ -6,17 +6,23 @@ import { AssetsAmount } from './AssetsAmount';
...
     expect(screen.getByText('Dai')).toBeInTheDocument();
   });
 
   it('should show positive values with plus', async () => {
-    render(<AssetsAmount amounts={MOCK_ASSETS_AMOUNTS} />);
-    expect(screen.getByText('14563943.834 ETH')).toBeInTheDocument();
+    render(
+      <AssetsAmount amounts={MOCK_ASSETS_AMOUNTS} title="Assets to Send" />
+    );
+    expect(screen.getByText('14.5 ETH')).toBeInTheDocument();
```
luizstacio, line 26: The test expected to show a + sign, but it didn't check it. Maybe change the value to include the +

packages/app/src/systems/DApp/hooks/useTxApprove.tsx
```diff
@@ -0,0 +1,77 @@
...
+  tx: (state: TxApproveMachineState) => state.context.tx,
+  receipts: (state: TxApproveMachineState) => state.context.receipts,
+  txDryRunError: (state: TxApproveMachineState) => state.context.txDryRunError,
+  isUnlockingLoading: (state: TxApproveMachineState) =>
+    state.children.unlock?.state.matches('unlocking'),
+};
+
+export function useTxApprove() {
+  const { account } = useAccount();
+  const service = useInterpret(() => txApproveMachine);
```
luizstacio, line 23:
```suggestion
  const service = useInterpret(txApproveMachine);
```

packages/app/src/systems/DApp/machines/txApproveMachine.tsx
```diff
@@ -0,0 +1,198 @@
...
+        invoke: {
+          src: 'approveTx',
+          data: {
+            input: (_: MachineContext, ev: { data: Wallet }) => {
+              return { tx: _.tx, wallet: ev.data };
+            },
+          },
+          onDone: [
+            {
+              target: 'done',
```
luizstacio, line 123: I think, we should show it the tx fails, and if it fails target failed.

packages/app/src/systems/DApp/machines/txApproveMachine.tsx
```diff
@@ -0,0 +1,198 @@
...
+            actions: ['assignTx'],
+          },
+        },
+      },
+      calculatingGas: {
+        invoke: {
+          src: 'calculateGas',
+          data: (_: MachineContext) => ({ input: { tx: _.tx } }),
+          onDone: [
+            {
```
luizstacio, line 67: It's possible to use the FetchMachine.errorState it should target failed maybe. Also, it already auto assign on the parent machine the field error with the error message. Removing the requirement of having the `assignTxDryRunError`

packages/app/src/systems/DApp/pages/TxApprove/TxApprove.tsx
```diff
@@ -1,77 +1,137 @@
...
-          onPress={handlers.approve}
-          isLoading={isLoadingTx}
-          isDisabled={isSent}
-        >
-          Send
-        </Button>
-      </Layout.BottomBar>
-    </Layout>
+          )}
+          {approvedTx && (
```
luizstacio, line 99: This could be a new page instead of a component. We should add a link on this content to open on the block-explorer.

packages/app/src/systems/DApp/routes.tsx
```diff
@@ -2,10 +2,11 @@ import { Route } from 'react-router-dom';
...
 import { Pages } from '../Core/types';
 
-import { SignatureRequest } from './pages';
+import { SignatureRequest, TxApprove } from './pages';
 
 export const dappRoutes = (
-  <Route path={Pages.signMessage()}>
-    <Route index element={<SignatureRequest />} />
+  <Route path={Pages.dapp()}>
+    <Route index path={Pages.signMessage()} element={<SignatureRequest />} />
```
luizstacio, line 9: Remove the index tag on both routes.

packages/app/src/systems/Transaction/utils/error.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+// typeguard
+const isVmErrorTypeKey = <K extends string & keyof GroupedError>(
+  properties: K[],
+  vmError: string
+): vmError is K => {
+  return (properties as string[]).includes(vmError);
+};
+
+export const getGroupedErrors = (rawErrors?: { message: string }[]) => {
```
luizstacio, line 34: Why we need this?

luizstacio (changes requested):

pedronauck (approved):

packages/app/src/systems/DApp/hooks/useTxApprove.tsx
```diff
@@ -0,0 +1,77 @@
...
+  const { send } = service;
+  const isUnlocking = useSelector(service, selectors.isUnlocking);
+  const isUnlockingLoading = useSelector(service, selectors.isUnlockingLoading);
+  const approvedTx = useSelector(service, selectors.approvedTx);
+  const tx = useSelector(service, selectors.tx);
+  const receipts = useSelector(service, selectors.receipts);
+  const txDryRunError = useSelector(service, selectors.txDryRunError);
+
+  const coinOutputs = getCoinOutputsFromTx(tx);
+  const outputsToSend = coinOutputs.filter(
```
pedronauck, line 33: This could be inside a `useMemo` since it's an array operation

pedronauck (changes requested):

packages/app/src/systems/Asset/components/AssetsAmount/AssetsAmount.test.tsx
LuizAsFight: we have `isPositive` and `isNegative` to show signals. I will add tests to it

When there's only `Assets to send`, We don't show the `-` signal to avoid confusion.

It only makes sense when you have both `Assets to send` and `Assets to receive`, then we use the signals to differentiate them.

packages/app/src/systems/DApp/machines/txApproveMachine.tsx
LuizAsFight: if you have more than one `fetchMachine` and wants to track the errors in different context variables, will not work.

that's why I prefer to use on this way.

but we should refact and move to `onFailed` as discussed before in slack channel

packages/app/src/systems/DApp/pages/TxApprove/TxApprove.tsx
LuizAsFight: it should send to "View Transaction" page, which is not done yet.

can we do this later when #38 gets done and merged?

LuizAsFight: I included the blockExplorer link for now

packages/app/src/systems/Transaction/utils/error.tsx
LuizAsFight: this file has some error helpers.

in `getGroupedErrors` it will return like:
image

then in the screens, we have the ability to add special treatments to any errors we want.
for example:
```
const { groupedErrors } = useTxApprove();

if (groupedErrors?.InsufficientInputAmount) {
  // do something specific to this error
}
```

packages/app/src/systems/Core/utils/provider.ts
```diff
@@ -0,0 +1,5 @@
+import { Provider } from 'fuels';
+
+import { VITE_FUEL_PROVIDER_URL } from '~/config';
+
+export const provider = new Provider(VITE_FUEL_PROVIDER_URL);
```
luizstacio, line 5: This should be on __mocks__ to avoid being used on other places of the applications, as the provider can be changed on the network selector.

packages/app/src/systems/DApp/machines/txApproveMachine.tsx
luizstacio: I think we should use the common field error and the `FetchMachine.errorState` to have a common pattern and we can move to onFailed after is refactored

luizstacio (changes requested):

pedronauck (approved):

LFSCamargo (approved):

luizstacio (approved):

PR #176, opened 2022-10-28
feat: amount input

### LINKED ISSUE #37: Create AmountInput component
## Figma Reference
https://www.figma.com/file/uEckQY4EaQ5rXwyISg6d1g/Fuel-Wallet?node-id=94%3A11474

## Notes
- This component will use `Input.Number` from our design system and use `rightElement` to add balance and max button

matt-user:
closes #37


LFSCamargo (dismissed):

pedronauck (changes requested):
CleanShot 2022-11-01 at 14 28 15@2x

Need to adjust the layout in the right elements of input @matt-user

packages/app/src/systems/Asset/components/AssetInput/AmountInput.tsx
```diff
@@ -0,0 +1,95 @@
+import { cssObj } from '@fuel-ui/css';
+import { Button, Flex, Input, Text } from '@fuel-ui/react';
+import type { BN } from 'fuels';
+import { useState } from 'react';
+import type { FC, ChangeEvent } from 'react';
+
+import { AmountInputLoader } from './AmountInputLoader';
+
+export type AmountInputProps = {
+  asset: {
```
pedronauck, line 10: You can use this instead for the `asset` prop:

```ts
amounts: Coin[] | TxOutputCoin[] | TxInputCoin[];
```

pedronauck (dismissed):

packages/app/src/systems/Asset/components/AssetInput/AmountInput.tsx
matt-user: This is the error I get when I try to change the type to your suggestion.  I get the same error on all of the types.
```
Property 'format' does not exist on type 'Coin | CoinTransactionRequestOutput | CoinTransactionRequestInput'.
  Property 'format' does not exist on type 'Coin'.
```

packages/app/src/systems/Asset/components/AssetInput/AmountInput.tsx
```diff
@@ -0,0 +1,100 @@
...
+};
+
+type AmountInputComponent = FC<AmountInputProps> & {
+  Loader: typeof AmountInputLoader;
+};
+
+export const AmountInput: AmountInputComponent = ({ amount }) => {
+  const [assetAmount, setAssetAmount] = useState<string>();
+  const formatArgs = {
+    minPrecision: 2,
```
LuizAsFight, line 20: should include 
`precision: DECIMAL_UNITS`,


otherwise more than 3 precision values won't be showed correctly

packages/app/src/systems/Asset/components/AssetInput/AmountInputLoader.tsx
```diff
@@ -0,0 +1,23 @@
+import { darkColors } from '@fuel-ui/css';
+import { CardList } from '@fuel-ui/react';
+import type { IContentLoaderProps } from 'react-content-loader';
+import ContentLoader from 'react-content-loader';
+
+export const AmountInputLoader = (props: IContentLoaderProps) => (
```
LuizAsFight, line 6: this loader is looking equal to `AssetItem` loader. which is wrong.

it should look similar to `AmountInput` component. look the video with both `AmountInput` and `AssetItem` examples:


https://user-images.githubusercontent.com/8636507/200440753-9cbfbc8a-b755-4104-a539-4ff3f1680f7d.mov

LuizAsFight (changes requested):

LuizAsFight:
Also, tests are broken. need to fix it

packages/app/src/systems/Asset/components/AssetInput/AmountInputLoader.tsx
matt-user: good catch!  I initially just took the loader from AssetItem and forgot to change it

packages/app/src/systems/Asset/components/AssetInput/AmountInput.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+export type AmountInputProps = {
+  amount: BN;
+};
+
+type AmountInputComponent = FC<AmountInputProps> & {
+  Loader: typeof AmountInputLoader;
+};
+
+export const AmountInput: AmountInputComponent = ({ amount }) => {
```
luizstacio, line 18: We are missing some props here to customize the component; From top of my head;
1. value (the amount on the input)
2. onChange (when the value is update we need to dispatch a func for the parent component)
3. balance (the balance that should be shown on the input)

You should also add this on the story book to see it working.

In the way it's made right now the amount with be set just once and the parent component will never know the new value.

luizstacio (changes requested):

packages/app/src/systems/Asset/components/AssetInput/AmountInput.tsx
matt-user: Just for clarification...
`value` is the displayed input in component?  ie what the user inputs into the component
onChange - this will replace the handleAmount change function?
balance - this is what is shown underneath the max button?  We get this value from the amount prop so I'm not sure why we would also pass in the balance

matt-user: I guess amount should probably be renamed to balance

matt-user: Thanks that makes sense I think I understand how to implement these changes

packages/app/src/systems/Asset/components/AssetInput/AmountInput.tsx
```diff
@@ -0,0 +1,109 @@
...
+export const AmountInput: AmountInputComponent = ({
+  balance,
+  value,
+  onChange,
+}) => {
+  const [assetAmount, setAssetAmount] = useState<string>(value);
+
+  const handleAmountChange = (event: ChangeEvent<HTMLInputElement>) => {
+    const newAssetAmount = handleAmountLeadingZeros(event);
+    onChange(event.target.value);
```
luizstacio, line 29:
```suggestion
    onChange(newAssetAmount);
```

luizstacio: We could change onChange and value to be and work with BN types.

luizstacio (changes requested):

packages/app/src/systems/Asset/components/AssetInput/AmountInput.tsx
```diff
@@ -0,0 +1,111 @@
...
+import { DECIMAL_UNITS, bn } from 'fuels';
+import type { BN } from 'fuels';
+import { useState } from 'react';
+import type { FC, ChangeEvent } from 'react';
+
+import { AmountInputLoader } from './AmountInputLoader';
+
+export type AmountInputProps = {
+  balance: BN;
+  value: BN;
```
LuizAsFight, line 12: I don't see why `value` prop should be an `BN`, you have a specific reason for that?

if not let's change to `value: string;`

it makes more sense because:
1- BN don't support floats , decimals.
2- BN is made to work with units, which is not the case here. You're working with regular decimals in this case
3- who's using the component, if got a value from API for example, will inform: `value={amount.format()}`, then the string already starts formatted here in the exact way the user inputed.

packages/app/src/systems/Asset/components/AssetInput/AmountInput.tsx
```diff
@@ -0,0 +1,111 @@
...
+  const handleAmountChange = (event: ChangeEvent<HTMLInputElement>) => {
+    const newAssetAmount = handleAmountLeadingZeros(event);
+    onChange(bn.parseUnits(event.target.value, DECIMAL_UNITS));
+    setAssetAmount(newAssetAmount);
+  };
+
+  const handlePress = () => {
+    onChange(balance);
+    setAssetAmount(balance.format({ precision: DECIMAL_UNITS }));
+  };
```
LuizAsFight, line 38: small logic change - comment 1/2

this 2 methods should use the same logic, having them separated opens space to bugs because someone can forget to update one of the sides.

I will suggest a commit but I didn't test it, it's just for you to understand the idea and use it if you thing makes sense

```suggestion
  const handleAmountChange = (amount: string) => {
    const newAssetAmount = handleAmountLeadingZeros(event);
    onChange(bn.parseUnits(newAssetAmount));
    setAssetAmount(newAssetAmount);
  };

  const handlePress = () => {
    handleAmountChange(balance.format({ precision: DECIMAL_UNITS }));
  };
```

packages/app/src/systems/Asset/components/AssetInput/AmountInput.tsx
```diff
@@ -0,0 +1,111 @@
...
+      <Input.Number
+        autoComplete="off"
+        inputMode="decimal"
+        name="amount"
+        placeholder="0.00"
+        allowedDecimalSeparators={['.', ',']}
+        allowNegative={false}
+        thousandSeparator={false}
+        value={assetAmount}
+        onChange={handleAmountChange}
```
LuizAsFight, line 63: small logic change - comment 2/2

```suggestion
        onChange={(e) => handleAmountChange(e.target.value)}
```

LuizAsFight (changes requested):
minor request

packages/app/src/systems/Asset/components/AssetInput/AmountInput.tsx
matt-user: I changed it bc of @luizstacio requested it in his above comment.  I may have misinterpreted his comment though.

packages/app/src/systems/Asset/components/AssetInput/AmountInput.tsx
matt-user: That makes sense thanks!

packages/app/src/systems/Asset/components/AssetInput/AmountInput.tsx
LuizAsFight: ok let's w8 for his answer

luizstacio: if you we don't abstract but pass just a string all the logic to handle the BN needs to be done on each screen. This is why I think receiving the BN is smarter on this case.

luizstacio (changes requested):
After setting the value via props, and trying to change the value on the input, the value is not set at first only on the second time.

https://user-images.githubusercontent.com/3941923/203431793-746600c4-a5a8-4dc3-aa1d-e0aed094e078.mov

packages/app/src/systems/Asset/components/AssetInput/AmountInput.tsx
LuizAsFight: @luizstacio it really makes sense to keep as value as `BN`.

`onChange` also sends a `BN` which should be consistent with `value` input

LuizAsFight (approved):

luizstacio (approved):

PR #185, opened 2022-11-01
feat: add release flow to wallet package

pedronauck:
## Tasks

- [x] Add changeset
- [x] Configure Github actions to release
- [x] Change package's name fo use `@fuel-wallet/*` instead of `@fuels-wallet/*`


.changeset/config.json
```diff
@@ -0,0 +1,18 @@
+{
+  "$schema": "https://unpkg.com/@changesets/config@2.0.0/schema.json",
+  "changelog": [
+    "@changesets/changelog-github",
+    { "repo": "FuelLabs/fuel-ts" }
```
luizstacio, line 5: This here should be `FuelLabs/fuel-wallet`?

.changeset/config.json
```diff
@@ -0,0 +1,18 @@
...
+  ],
+  "commit": false,
+  "fixed": [["@fuel-wallet/*"]],
+  "linked": [],
+  "access": "public",
+  "baseBranch": "master",
+  "updateInternalDependencies": "patch",
+  "ignore": ["app", "page"],
```
luizstacio, line 13: Should we ignore, store, config, and fuelhat? The idea is to release just `sdk` and `types` for now.

luizstacio (changes requested):

.changeset/config.json
pedronauck: ✅

LuizAsFight (dismissed):

.changeset/config.json
pedronauck: ✅

luizstacio (dismissed):

luizstacio (approved):

LuizAsFight (approved):

PR #187, opened 2022-11-02
feat(app): export seed phrase screen

LFSCamargo:
- Adds export vaults service
- Adds the settings machine
- Adds tests to the settings machine
- Adds a new section on the drawer
- Adds a new screen to the app

<div style="display: flex;">

</div>


packages/app/src/systems/Account/services/account.ts
```diff
@@ -149,7 +156,19 @@ export class AccountService {
     // TODO: fix this on fuel-ts it should be possible to
     // customize the ProviderURL on the manager level
     wallet.provider = new Provider(VITE_FUEL_PROVIDER_URL);
-    return wallet;
+
+    return {
+      ...(wallet as WalletUnlocked),
+      exportVault() {
```
luizstacio, line 162: I think we should implement this in another way to strictly expose mnemonic only when it's needed.

Maybe passing some configs when spawning the unlock machine.

luizstacio (changes requested):

pedronauck (changes requested):
All review considerations I made [on this pull request](https://github.com/FuelLabs/fuels-wallet/pull/188) are the same here, since both have same code.

PR #197, opened 2022-11-12
feat(docs): new sdk documentation template

pedronauck:
## Tasks

- [x] Setup NextJS with MDX to auto-generate pages
- [x] Integrate DocSearch
- [x] Auto scrap Docsearch using Github Action
- [x] Deploy on Vercel
- [x] Fix `@fuel-ui` NextJS integration


luizstacio (changes requested):
- [x] Increase contrast on texts to improve readability
- [x] Add the video on install instructions
- [x] Fix layout responsiveness  
- [ ] Replace code snippets on the md with real code snippets
   - [ ] Move the code examples to separated files
   - [ ] Move the tests of the code snippets to the side of this files
   - [ ] Add the code snippet import on the markdown files.

.github/workflows/release.yaml
```diff
@@ -33,7 +33,7 @@ jobs:
       - name: Bump and Collect Version
         run: |
           pnpm changeset version
-          echo "BUILD_VERSION=$(sed -nE 's/^\s*"version": "(.*?)",$/\1/p' design-system/react/package.json)" >> $GITHUB_ENV
+          echo "BUILD_VERSION=$(sed -nE 's/^\s*"version": "(.*?)",$/\1/p' packages/app/package.json)" >> $GITHUB_ENV
```
luizstacio, line 36: We should split this fixes in small PRs: https://github.com/FuelLabs/fuels-wallet/pull/255

.gitignore
```diff
@@ -1,6 +1,8 @@
 # Generated by Cargo
 # will have compiled files and executables
 /target/
+/out
+packages/docs/.vscode
```
luizstacio, line 5: Do we need this?

packages/docs/.env.example
```diff
@@ -0,0 +1,6 @@
+NEXT_PUBLIC_ALGOLIA_APP_ID=
+NEXT_PUBLIC_ALGOLIA_SEARCH_API_KEY=
+ALGOLIA_SEARCH_ADMIN_KEY=
```
luizstacio, line 3: We can remove `ALGOLIA_SEARCH_ADMIN_KEY` from the example as it is used only on the CI

luizstacio (changes requested):

scripts/gh-pages-preview.mjs
```diff
@@ -0,0 +1,15 @@
+import { join } from "path";
```
luizstacio, line 1: Just as a note, this refactoring could be done on a different PR from the docs PR.

packages/docs/src/lib/str.ts
```diff
@@ -0,0 +1,3 @@
+export function capitalize(str: string) {
+  return `${str.slice(0, 1).toUpperCase()}${str.slice(1, Infinity)}`;
```
luizstacio, line 2: Why pass Infinity here?
```suggestion
  return `${str.slice(0, 1).toUpperCase()}${str.slice(1)}`;
```

luizstacio (changes requested):

.github/workflows/release.yaml
pedronauck: I just fix because I saw it

.gitignore
pedronauck: Next13 it's auto-generating this automatically, I could see how to disable this option. So, I just ignored it.

packages/docs/.env.example
pedronauck: ✅

scripts/gh-pages-preview.mjs
pedronauck: Noted

packages/docs/src/lib/str.ts
pedronauck: I didn't know that by omitting the second parameter it got it the rest, cool!

luizstacio (approved):

luizstacio (approved):

PR #287, opened 2022-11-30
feat: persist balance visibility after page reload

### LINKED ISSUE #119: Hide balance icon should be persisted
As a user, when clicking to hide my balance, is expected that the next time I open the wallet, the balance will be yet hidden.

Today the state is reset on every refresh.

Screen Shot 2022-09-23 at 1 02 43 AM

tomiiide:
## What does this PR do?
Fixes #119 
This PR makes the Wallet Balance Visibility persistent on page reload.


packages/app/src/systems/Account/components/BalanceWidget/BalanceWidget.tsx
```diff
@@ -12,6 +12,8 @@ import type { Account } from '@fuel-wallet/types';
 import { useEffect, useState } from 'react';
 import type { ReactNode } from 'react';
 
+import type { AccountInputs } from '../..';
```
luizstacio, line 15: inside the components, you should import from the components folder to avoid cycle dependencies.

packages/app/src/systems/Account/components/BalanceWidget/BalanceWidget.tsx
```diff
@@ -87,8 +100,8 @@ export function BalanceWidget({
           <VisibilityButton
             aria-label={isHidden ? 'Show balance' : 'Hide balance'}
             isHidden={isHidden}
-            onHide={() => setIsHidden(true)}
-            onShow={() => setIsHidden(false)}
+            onHide={() => toggleBalanceVisibility(true)}
```
luizstacio, line 103: The name of the function was better before, toggle would be better for a function that doesn't receive the state. But in this case, we explicitly set is visible or not.

```suggestion
            onHide={() => setBalanceVisibility(true)}
```

luizstacio (changes requested):
We should add some tests for this action.

LuizAsFight (approved):

LuizAsFight (changes requested):
code looks good, but tests are failing, need to fix it

packages/app/src/systems/Account/components/BalanceWidget/BalanceWidget.tsx
```diff
@@ -34,19 +35,32 @@ export type BalanceWidgetProps = {
   account?: Account;
   isHidden?: boolean;
   isLoading?: boolean;
+  onSetBalanceVisibility?: (
```
luizstacio, line 38:
```suggestion
  onChangeBalanceVisibility?: (
```

luizstacio (changes requested):

packages/app/src/systems/Account/components/BalanceWidget/BalanceWidget.tsx
```diff
@@ -14,6 +14,7 @@ import type { ReactNode } from 'react';
 
 import { BalanceWidgetLoader } from './BalanceWidgetLoader';
 
+import type { AccountInputs } from '~/systems/Account';
```
luizstacio, line 17:
```suggestion
import type { AccountInputs } from '~/systems/Account/services';
```

luizstacio (changes requested):

luizstacio (approved):

PR #309, opened 2022-12-02
feat: add TxStatusAlert

### LINKED ISSUE #303: Create TxStatusAlert
### Figma Reference
https://www.figma.com/file/uEckQY4EaQ5rXwyISg6d1g/Fuel-Wallet?node-id=290%3A15938&t=dMDQpq12twuDJmlp-0

It should have this states: 
- Failed Transaction
- Pending Transaction
- Invalid TX ID
- Transaction Not Found

LuizAsFight:
Closes #303


packages/app/src/systems/Transaction/components/TxStatusAlert/TxStatusAlert.tsx
```diff
@@ -0,0 +1,101 @@
...
+            <Link
+              aria-label="View Transaction on Block Explorer"
+              color={txColor}
+              isExternal
+              href={getBlockExplorerLink({
+                path: `transaction/${txId}`,
+                providerUrl,
+              })}
+              css={{
+                '&:focus-visible': {
```
pedronauck, line 60: This should be inside the `styles` object as well

packages/app/src/systems/Transaction/components/TxStatusAlert/TxStatusAlert.tsx
```diff
@@ -0,0 +1,101 @@
...
+  providerUrl,
+}) => {
+  const alertStatus = useMemo(() => {
+    if (txStatus === TxStatus.pending) return 'warning';
+    if (txStatus === TxStatus.success) return 'success';
+    if (txStatus === TxStatus.error || error) return 'error';
+
+    return 'info';
+  }, [txStatus, error]);
+  const txColor = error
```
pedronauck, line 30: Some line break here is a good idea 😅

pedronauck (changes requested):

packages/sdk/src/utils/blockExplorer.ts
```diff
@@ -6,9 +6,9 @@ export function getBlockExplorerLink({
...
 }: {
   path: string;
-  providerUrl: string;
+  providerUrl?: string;
 }) {
-  return `${BLOCK_EXPLORER_URL}${path}?providerUrl=${encodeURIComponent(
-    providerUrl
-  )}`;
+  return `${BLOCK_EXPLORER_URL}${path}${
+    providerUrl ? `?providerUrl=${encodeURIComponent(providerUrl)}` : ''
```
luizstacio, line 12: I think we should not make `providerUrl` optional as we want to enforce we are passing it when opening block explorer.

luizstacio (changes requested):

pedronauck (approved):

packages/sdk/src/utils/blockExplorer.ts
LuizAsFight: - this is a util function for the whole app, it should accept optional and send to default provider. it's up
- this function is just temporary, we're gonna use from SDK when [fuels-ts/309](https://github.com/FuelLabs/fuels-wallet/pull/309) gets merged


if you think `providerUrl` should be always informed, [this](https://github.com/FuelLabs/fuels-wallet/pull/309) is the right place to ask for it

luizstacio: On SDK level it should not be required but here yes

luizstacio (approved):

PR #317, opened 2022-12-02
feat: add fuelweb on deps

luizstacio:
When FuelWeb3 was loaded all places that useLoading were not receiving the new FuelWeb3 reference.


tomiiide (approved):
Looks good to me!

PR #324, opened 2022-12-03
feat(app): add send feature

pedronauck:
## Tasks

- [x] Adapt components from `Transaction` and `DApp/TransactionRequest` to be able to re-use them
- [x] Add missing components no core module 
- [x] Create all transactions-related screens and feature

https://user-images.githubusercontent.com/2029172/209880638-93d75732-0532-4812-803d-d2c4488780d0.mp4


packages/app/src/systems/Core/utils/machine.ts
```diff
@@ -1,5 +1,6 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
-import type { AnyInterpreter, StateFrom, Action } from 'xstate';
+import * as R from 'rambda/immutable';
```
luizstacio, line 2: We should avoid using new libs when they are not needed. also, rambda even been good. It Would be better to have a more known lib or just write the helpers functions needed in this case.

packages/app/src/systems/Send/machines/sendMachine.ts
```diff
@@ -0,0 +1,298 @@
...
+    actions: {
+      reset: resetContext(),
+      assignAsset: assignWithInput('inputs.asset'),
+      assignAddress: assignWithInput('inputs.address'),
+      assignAmount: assignWithInput('inputs.amount'),
+      assignWallet: assignWithData('inputs.wallet'),
+      assignTxRequest: assignWithData('response', true),
+      assignTxApprove: assignWithData('response', true),
+      assignTxRequestError: assignErrors('txRequestError'),
+      assignTxApproveError: assignErrors('txApproveError'),
```
luizstacio, line 233: This looks like a great abstraction, but we should avoid including it in this PR as it will create a new pattern only used here. Looking at the feature, this kind of thing can create many machines without the same patterns. A better approach is to split the PR into phases.

1. Add the new abstraction
2. Refactor all the machines to use the new abstraction
3. Use the new abstraction here, as all the code base is updated.

luizstacio (changes requested):

packages/app/src/systems/Send/machines/sendMachine.ts
pedronauck: Can I let this PR with this modifications and then open a new one including them in the other machines @luizstacio ?

packages/app/src/systems/Core/utils/machine.ts
pedronauck: ✅

packages/app/src/systems/Send/machines/sendMachine.ts
luizstacio: In my opinion, we should focus on refactoring it first and after add new abstractions. When doing abstractions we create new patterns to implement things that make current code that can be done or in progress already legacy. A good approach for this kind of thing would be to create an issue and do when is a good time. This would solve the two things as all the code base will follow the pattern, and after all, the code would be refactored.

But I want another opinion from the team, not to be the single arbiter here. @LuizAsFight

LuizAsFight: I agree. We should not merge this part because:
- it's orthogonal to the main objective of the issue: Create send transaction screen
- we should not just implement a whole new pattern when we already have ~10 machines created in other way


Could you open one issue and elaborate what will be the benefits of refactor to include a new layer to manage actions, with a bunch of helpers and complex object management? then we can continue talking from there

LuizAsFight:
I'm reviewing this PR code already, but will first mention functional problems I found

Send Screen:
- in first frame of video, we see a blink of “ConfirmSend” screen before showing unlock dialog
- can we ask password to unlock wallet later in the flow? we only need signature for actually sending the transaction right @luizstacio ?  
- transaction errors happening, but no feedback to the user.
  - we should not try to simulate the transaction if we know that:
    - invalid address
      - validate To field using  isB256 and isBech32 methods exposed by TS SDK
    - amount bigger than balance
      - validate InputAmount
      - if we click on “Max”, balance is set but considering fee. can we fake a transaction as soon as the user access the screen, enabling us to show the transaction fee earlier and cover all scenarios?
    - anyway after validation, if error occurs when trying to simulate, we might want to show TxErrors component as feedback
- regarding Transaction Details card.
  - should be always visible with value of faked transaction mentioned in prev topic
  - if we won’t do this transaction fee predict, at least it should be hidden when inputs get clear after getting transaction. Right now it’s still visible showing 0.000 ETH fee.


here's a record showing all topics mentioned:

https://user-images.githubusercontent.com/8636507/207742791-c3603ee2-c192-49eb-a556-01e9eecbee54.mov

packages/app/src/systems/Send/machines/sendMachine.ts
```diff
@@ -0,0 +1,298 @@
...
+      },
+      SET_AMOUNT: {
+        actions: ['assignAmount'],
+        target: 'idle',
+      },
+      RESET: {
+        actions: ['reset'],
+        target: 'idle',
+      },
+    },
```
LuizAsFight, line 221: all this events are global in the machine?

looks like they are specific of `idle` state? if yes let's move them to right state

packages/app/src/systems/Send/machines/sendMachine.ts
```diff
@@ -0,0 +1,298 @@
...
+            actions: [
+              send<MachineContext, UnlockMachineEvents>(
+                (_, ev) => ({ type: 'UNLOCK_WALLET', input: ev.input }),
+                { to: 'unlock' }
+              ),
+            ],
+          },
+        },
+      },
+      idle: {
```
LuizAsFight, line 120: this `idle` is not actually `idle` as it does a lot of things in substates...

should be same style of `confirming` state, that has a first `idle` substate

LuizAsFight (changes requested):

packages/app/src/systems/Asset/components/AssetSelect/AssetSelect.stories.tsx
```diff
@@ -0,0 +1,26 @@
...
+  const [selected, setSelected] = useState<AssetSelectInput>(null as any);
+  return (
+    <Box css={{ maxWidth: 300 }}>
+      <AssetSelect
+        selected={selected}
+        onSelect={(asset) => setSelected(asset!)}
+        items={MOCK_ASSETS_AMOUNTS}
+      />
+    </Box>
+  );
```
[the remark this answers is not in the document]
pedronauck: ✅

packages/app/src/systems/Asset/components/AssetSelect/AssetSelect.test.tsx
```diff
@@ -0,0 +1,65 @@
...
+  return (
+    <AssetSelect
+      items={MOCK_ASSETS_AMOUNTS}
+      selected={selected}
+      onSelect={handleSelect}
+    />
+  );
+}
+
+describe('AssetSelect', () => {
```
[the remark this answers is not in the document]
pedronauck: ✅

packages/app/src/systems/Transaction/components/TxErrors/TxErrors.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+const styles = {
+  root: cssObj({
+    padding: '$3 $4',
+
+    '.fuel_alert--content': {
+      gap: '$1',
+    },
+  }),
+};
```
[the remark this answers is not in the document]
pedronauck: ✅

packages/app/src/systems/Transaction/components/TxDetails/TxDetails.tsx
```diff
@@ -16,10 +16,11 @@ type TxDetailsComponent = FC<TxDetailsProps> & {
 };
 
 export const TxDetails: TxDetailsComponent = ({
-  fee,
+  fee: initialFee,
   amountSent,
 }: TxDetailsProps) => {
-  const total = fee?.add(bn(amountSent));
+  const fee = bn(initialFee || 0);
```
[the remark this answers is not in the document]
pedronauck: ✅

packages/app/src/systems/Transaction/components/TxDetails/TxDetails.tsx
```diff
@@ -31,7 +32,7 @@ export const TxDetails: TxDetailsComponent = ({
             <Flex css={styles.detailItem}>
               <Text as="span">Fee (network)</Text>
               <Text as="span" aria-label="Gas Value">
-                {fee?.format()} ETH
+                {fee?.format({ precision: DECIMAL_UNITS })} ETH
```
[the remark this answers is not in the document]
pedronauck: ✅

packages/app/src/systems/Home/components/HomeActions/HomeActions.tsx
```diff
@@ -12,11 +14,27 @@ export const HomeActions = ({
...
+    <Button
+      aria-label="Send Button"
+      onPress={sendAction}
+      isDisabled={isDisabled || !hasBalance}
+      css={styles.button}
+      isLoading={isLoading}
+    >
+      Send
+    </Button>
+  );
```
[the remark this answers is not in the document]
pedronauck: because right after we need to check `hasBalance` to show it wrapped in a `Tooltip` or not, so to don't repeat the component entirely inside JSX, we can just save it inside a variable and do this:

```jsx
{hasBalance ? (
  sendButton
) : (
  <Tooltip content="You don't have balance to send">{sendButton}</Tooltip>
)}
```

packages/app/src/systems/Send/routes.tsx
```diff
@@ -0,0 +1,7 @@
+import { Route } from 'react-router-dom';
+
+import { Pages } from '../Core/types';
+
+import { Send } from './pages';
+
+export const sendRoutes = <Route path={Pages.send()} element={<Send />} />;
```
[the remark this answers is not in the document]
pedronauck: It's defined as a "section" inside the design. It has a bunch of functionality, components, hooks, etc. The definition of a system isn't tight of where it's, but how it's, how complex and the ability to be able to use this system coupled of other parts of the application. All these specs fit very well for the Send.

packages/app/src/systems/Send/components/SendConfirm/SendConfirm.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+export function SendConfirm({ inputs, response, errors }: SendConfirmProps) {
+  const { outputsToSend } = useTxOutputs(response?.txRequest);
+  return (
+    <Stack gap="$3" css={styles.root}>
+      <Stack as="header" gap="$1">
+        <Heading as="h3">Confirm before send</Heading>
+        <Text>
+          Check your transaction summary and confirm all details is right before
+          send
```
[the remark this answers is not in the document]
pedronauck: ✅

packages/app/src/systems/Send/components/SendConfirm/SendConfirm.tsx
```diff
@@ -0,0 +1,55 @@
...
+import { TxDetails, TxFromTo, useTxOutputs } from '~/systems/Transaction';
+
+type SendConfirmProps = UseSendReturn;
+
+export function SendConfirm({ inputs, response, errors }: SendConfirmProps) {
+  const { outputsToSend } = useTxOutputs(response?.txRequest);
+  return (
+    <Stack gap="$3" css={styles.root}>
+      <Stack as="header" gap="$1">
+        <Heading as="h3">Confirm before send</Heading>
```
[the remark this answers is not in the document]
pedronauck: ✅

packages/app/src/systems/Send/components/SendSelect/SendSelect.tsx
```diff
@@ -0,0 +1,86 @@
...
+            onChange={(e) => send.handlers.setAddress(e.target.value)}
+          />
+        </Input>
+      </Flex>
+      <Stack gap="$3">
+        <Text as="span" css={{ ...styles.title, ...styles.amountTitle }}>
+          Which amount?
+        </Text>
+        <InputAmount
+          value={send.inputs?.amount || bn(0)}
```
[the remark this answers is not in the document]
pedronauck: ✅

packages/app/src/systems/Send/components/SendSelect/SendSelect.tsx
```diff
@@ -0,0 +1,86 @@
...
+          onChange={send.handlers.setAmount}
+          balance={bn(tx.account?.balance || 0)}
+        />
+      </Stack>
+      {send.response?.fee && (
```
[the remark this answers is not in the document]
pedronauck: ✅

packages/app/src/systems/Send/machines/sendMachine.ts
pedronauck: Nope, they need to be global, because if the machine are in the `checking` state getting info about transaction, and the user start to change some of the form values, the machine need to go back to the `idle`

packages/app/src/systems/Send/machines/sendMachine.ts
pedronauck: ✅

luizstacio (changes requested):
I think just move out of this PR the utility tools for the machine.

packages/docs/package.json
```diff
@@ -5,7 +5,7 @@
   "scripts": {
```
luizstacio, line 5: We should remove the changes from "export" and "build/utils" from this PR as they are orthogonal changes.

luizstacio: Also the preview deployment is broken on the current PR.

.changeset/warm-chefs-begin.md
```diff
@@ -0,0 +1,2 @@
+---
```
luizstacio, line 1: We can delete this file as not changes were made to the SDK or types packages.

luizstacio (changes requested):
The current preview is broken on this PR.
Also will be good to hold to use the changes of the PR #365

package.json
```diff
@@ -51,7 +51,7 @@
     "test:ci": "pnpm test -- --ci --testLocationInResults --json --coverage",
     "test:clear": "pnpm -r exec jest --clearCache",
     "test:coverage": "pnpm run --filter=app test --coverage",
-    "test:e2e": "playwright test --config=packages/app/playwright.config.ts",
+    "test:e2e": "NODE_ENV=test playwright test --config=packages/app/playwright.config.ts",
```
tomiiide, line 54: Thank you for this.

packages/app/src/systems/Send/machines/sendMachine.ts
LuizAsFight: interesting

luizstacio (changes requested):
I'm first evaluating just the flow;
1. When typing the destination address, the console throws errors that should not be displayed.
Screenshot 2022-12-28 at 00 23 37
2. When clicking to confirm, it goes throw a loading state. This shows the current screen. We should go to a transaction view screen if the correct loaders.
Screenshot 2022-12-28 at 00 24 02
3. The confirmation page should re-use the transaction request screen or its components. The current screen needs to include information like gas fees.
Screenshot 2022-12-28 at 00 24 13
4. After approving the transaction, we can open the [review transaction screen](https://fuels-wallet-gdaskxprw-fuel-labs.vercel.app/storybook/?path=/story/transaction-pages-txview--usage), removing the need for the custom status report screens. 

On the current request transaction flow, the screen loading and review for approval are shown like this;
Screenshot 2022-12-28 at 00 30 56
Screenshot 2022-12-28 at 00 29 02

pedronauck:
@luizstacio I did all changes you mentioned in the last comment _(the video above is simulating slow 3g network)_

https://user-images.githubusercontent.com/2029172/209875784-5257c88d-991b-4627-a2d7-09dbaaf0dbe6.mp4

luizstacio:
> @luizstacio I did all changes you mentioned in the last comment _(the video above is simulating slow 3g network)_
> 
>  Kapture.2022-12-28.at.18.44.58.mp4

Perfect just two more things;
- After approval and sending, the header could be titled sending the transaction.
- On the transaction view after sending, the status is not green, but the tx is already confirmed.
Screenshot 2022-12-30 at 01 28 12

luizstacio (changes requested):
Perfect just two more things;
- After approval and sending, the header could be titled sending the transaction.
- On the transaction view after sending, the status is not green, but the tx is already confirmed.
   Screenshot 2022-12-30 at 01 28 12
- Remove the images that are not needed anymore.

luizstacio (commented):
We should display the address using bech32 here, maybe not related to the PR;
Screenshot 2023-01-09 at 20 55 14

packages/app/src/systems/Asset/components/AssetSelect/AssetSelect.tsx
```diff
@@ -0,0 +1,243 @@
...
+      <Dropdown.Menu
+        autoFocus
+        aria-label="Actions"
+        css={styles.menu(width)}
+        onAction={handleSelect}
+      >
+        {(items || []).map((item) => {
+          const asset = getAssetInfoById(item.assetId, item);
+          const amount = bn(asset.amount);
+          const amountStr = `${amount.format({
```
luizstacio, line 121: Use `formatAmount` here.

packages/app/src/systems/Asset/components/AssetSelect/AssetSelect.tsx
```diff
@@ -0,0 +1,243 @@
...
+  selected?: Maybe<AssetSelectInput>;
+};
+
+export function AssetSelect({
+  items,
+  selected,
+  onSelect,
+  ...props
+}: AssetSelectProps) {
+  /** i'm using a ref here instead of a state in order to don't trigger a new re-render */
```
luizstacio, line 39: I think we don't need this comment here?

luizstacio (changes requested):

packages/app/src/systems/Asset/components/AssetSelect/AssetSelect.tsx
```diff
@@ -0,0 +1,243 @@
...
+  function handleSetWidth(isOpen: boolean) {
+    if (isOpen) {
+      const item = document.querySelector('#fuel_asset-select');
+      setWidth(item?.clientWidth ?? null);
+    }
+    /**
+     * I did this using dom manipulation instead of useState in order
+     * to avoid an unnecessary re-render
+     */
+    const arrow = document.querySelector('#fuel_asset-select > .fuel_icon');
```
luizstacio, line 61: Even with state changes is better to keep using react than access dom.

packages/app/src/systems/Asset/components/AssetSelect/AssetSelect.tsx
```diff
@@ -0,0 +1,243 @@
...
+};
+
+export function AssetSelect({
+  items,
+  selected,
+  onSelect,
+  ...props
+}: AssetSelectProps) {
+  /** i'm using a ref here instead of a state in order to don't trigger a new re-render */
+  const [width, setWidth] = useState<Maybe<number>>(null);
```
luizstacio, line 40: I think we should keep the menu with a hard width, reducing the new for dynamic control here. Any issues on the component that requires us to use dynamic width?

packages/app/src/systems/Asset/components/AssetSelect/AssetSelect.tsx
```diff
@@ -0,0 +1,243 @@
...
+    /**
+     * I did this using dom manipulation instead of useState in order
+     * to avoid an unnecessary re-render
+     */
+    const arrow = document.querySelector('#fuel_asset-select > .fuel_icon');
+    arrow?.classList.toggle('rotate');
+  }
+
+  return (
+    <Dropdown
```
luizstacio, line 66: Maybe break the component in smaller components?

packages/app/src/systems/DApp/components/index.tsx
```diff
@@ -1 +1,2 @@
 export * from './ConnectInfo';
+export * from '../../Transaction/components/TxContent';
```
luizstacio, line 2: Maybe we need to fix the imports here, as it looks bad to export on index a component from other place.

luizstacio (changes requested):

packages/app/src/systems/Send/hooks/useSend.tsx
```diff
@@ -0,0 +1,174 @@
...
+        if (state.matches('fetchingFakeTx')) return SendStatus.loading;
+        if (isLoadingTx) return SendStatus.loadingTx;
+        if (state.matches('confirming')) return SendStatus.confirming;
+        return SendStatus.selecting;
+      },
+      [txStatus]
+    );
+  },
+  title(txStatus?: TxRequestStatus) {
+    return useCallback(
```
luizstacio, line 50: xstate service already caches the answer. Why do we need to use callback here?

luizstacio (changes requested):

LuizAsFight:
functional when running locally

- we have this "suggested" address, is this intended?
  - if yes it's a dev only thing?
- final of amount input is cut (notice Max button)

image

LuizAsFight:
https://user-images.githubusercontent.com/8636507/211689475-66efa75c-b970-41c8-b74a-81dda0ed3643.mov

can we validate if amount is bigger than balance before sending the transaction? 
and show feedback similar to 
image

LuizAsFight:
storybook Send page is broken in PR`s preview

image

packages/app/src/systems/Transaction/services/transaction.tsx
```diff
@@ -1,12 +1,31 @@
...
+  Provider,
+  TransactionResponse,
+} from 'fuels';
 
 import type { Transaction } from '../types';
 import { toJSON } from '../utils';
 
+import { ASSET_MAP, isEth } from '~/systems/Asset';
 import { db, uniqueId } from '~/systems/Core';
+import { provider } from '~/systems/DApp/__mocks__/dapp-provider';
```
LuizAsFight, line 28: I think we should not use things from mock file inside a service, would prefer to just create the `provider` here

packages/app/src/systems/Transaction/services/transaction.tsx
```diff
@@ -98,7 +139,88 @@ export class TxService {
...
+      );
+      request.addResources(resources);
+    }
+    return request;
+  }
+
+  static async fundTransaction(input: TxInputs['fundTransaction']) {
+    const { wallet, tx } = input;
+    const preRequest = await TxService.getTxWithResources({ wallet, tx });
+    const txCost = await wallet.provider.getTransactionCost(preRequest);
```
LuizAsFight, line 178: can we try to get the same `txCost` info, but using `tx utils` ? would be better because:

1- `getTransactionCost` is broken, because it uses `calculateTransactionFee`, which is using a fixed `GAS_PRICE_FACTOR`.
2- `tx utils` will become the default way to get this kind of info, and this provider methods will be deprecated

packages/app/src/systems/Transaction/services/transaction.tsx
```diff
@@ -98,7 +139,88 @@ export class TxService {
...
+  static async createFakeTx() {
+    const toWallet = Wallet.generate();
+    const params = { gasLimit: MAX_GAS_PER_TX };
+    const request = new ScriptTransactionRequest(params);
+    const dest = toWallet.address;
+    const assetId = Object.values(ASSET_MAP)[0].assetId;
+    const amount = bn(1);
+
+    request.addCoinOutput(dest, amount, assetId);
+    const fee = request.calculateFee();
```
LuizAsFight, line 203: can we try to get the same `fee` info, but using `tx utils` ? would be better because:

1- `calculateFee` is broken, because it's using a fixed `GAS_PRICE_FACTOR`.
2- `tx utils` will become the default way to get this kind of info, and this provider methods will be deprecated

packages/docs/examples/Transfer.tsx
```diff
@@ -22,30 +12,36 @@ import { useLoading } from '~/src/hooks/useLoading';
 export function Transfer() {
   const [fuel, notDetected] = useFuel();
   const [isConnected] = useIsConnected();
-  const [txId, setTxId] = useState<string>('');
   const [amount, setAmount] = useState<BN>(bn.parseUnits('0.00001'));
+  const [addr, setAddr] = useState<string>(
+    'fuel1a6msn9zmjpvv84g08y3t6x6flykw622s48k2lqg257pf9924pnfq50tdmw'
+  );
```
LuizAsFight, line 18: 👏🏻

LuizAsFight (changes requested):

luizstacio (changes requested):
Balance is not refreshing after the send is complete

luizstacio (dismissed):

packages/app/src/systems/Send/machines/sendMachine.ts
```diff
@@ -0,0 +1,160 @@
...
+      assignInitialFee: assign({
+        fee: (_, ev) => ev.data,
+      }),
+    },
+    services: {
+      fetchFakeTx: FetchMachine.create<null, BN>({
+        showError: false,
+        async fetch() {
+          const tx = await TxService.createFakeTx();
+          return tx.request.calculateFee().amount;
```
LuizAsFight, line 123: can we use tx utils to get the fee instead?

packages/app/src/systems/DApp/__mocks__/dapp-transaction.ts
```diff
@@ -14,14 +14,14 @@ export const getMockedTransaction = async (
...
+  const transactionRequest = new ScriptTransactionRequest(params);
   const destinyAddress = Address.fromPublicKey(destiny);
   const ownerAddress = Address.fromPublicKey(owner);
   const assetId =
     '0x0000000000000000000000000000000000000000000000000000000000000000';
   const amount = bn.parseUnits('0.1').toNumber();
-  txRequest.addCoinOutput(destinyAddress, amount, assetId);
-  const fee = txRequest.calculateFee();
+  transactionRequest.addCoinOutput(destinyAddress, amount, assetId);
+  const fee = transactionRequest.calculateFee();
```
LuizAsFight, line 24: can we use tx utils to get the fee instead?

LuizAsFight (changes requested):

LuizAsFight (approved):

PR #335, opened 2022-12-06
fix: service worker path

luizstacio:
Fix mockServiceWorker path for storybook.
Also, find a issue on `Home` page of the storybook https://github.com/FuelLabs/fuels-wallet/issues/336

### DIFF
--- a/.changeset/flat-games-melt.md
+++ b/.changeset/flat-games-melt.md
@@ -1,0 +1,5 @@
+---
+"@fuel-wallet/sdk": patch
+---
+
+chore: allow option providerUrl in getBlockExplorer function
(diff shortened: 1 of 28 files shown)


LuizAsFight (changes requested):
still not working when testing in `preview` storybook...


https://user-images.githubusercontent.com/8636507/205938509-a9cf871e-a488-473e-aaf8-e359939d482e.mov

- `No Assets` and `With Assets` pages are never leaving loading state
- `With Assets` should show balance

luizstacio:
> still not working when testing in `preview` storybook...
> 
>  Screen.Recording.2022-12-06.at.11.27.19.mov 
> * `No Assets` and `With Assets` pages are never leaving loading state
> * `With Assets` should show balance

I open an issue to track it that is on description #336. We can move with this one and fix this one later.

pedronauck (approved):
Lgfm

LuizAsFight (approved):

Issue #281, opened 2022-11-30
Bug with seed phrase if type it manually


cepdy4ek27:
### What version of Wallet are you using?

0.0.1

### What is happening?

While registering a wallet, if I enter the seed phrase manually, then at the 12th word the system does not wait until I end it and gives an error that the phrase is not correct. After that - the page is updated and then I see a gray screen with no buttons.

If use Copy-Paste for 12 words at the beggining - it works correct

### steps to reproduce

1. Create new wallet
2. Try to type all 12 words of seed phrase manually, if not use Copy-paste function.
3. On the 12th word, system don't give you to finish your 12th word and displayed error. 
4. The page is automatically updated and user get gray screen with no buttons.

### How it should behave?

It should behave as always on other wallets. Type 12 words, button "Confirm".

### What os are you seeing the problem on?

Mac/Apple Desktop

### What browser?

Chrome

### What is your browser version?

107.0.5304.110

luizstacio:
Thanks for the report, I have tested it and was able to emulate it.

https://user-images.githubusercontent.com/3941923/205221831-681b67c2-e938-4675-a694-b9c631e21893.mp4

PR #341, opened 2022-12-06
fix: when enter typing the mnemonic the setup page breaks

luizstacio:
The issue was related to accepting any mnemonic without checking if it was valid or not, making the page break when the mnemonic was invalid.

For this, I also add two feedbacks for users now, where;
- When the mnemonic is invalid, it shows a message for invalid;
- When the mnemonic does not match, the generated one shows a message for the seed phrase not match.

closes: #281


packages/app/src/systems/SignUp/machines/signUpMachine.ts
```diff
@@ -86,20 +90,48 @@ export const signUpMachine = createMachine(
         },
       },
       waitingMnemonic: {
+        initial: 'invalidMnemonic',
+        states: {
+          invalidMnemonic: {
+            entry: [
+              assignErrorMessage(
+                'The seed phrase is not valid check the words for typos or missing words'
```
LuizAsFight, line 98:
```suggestion
                'The seed phrase is not valid. Check the phrase for typos or missing words'
```

packages/app/src/systems/SignUp/machines/signUpMachine.ts
```diff
@@ -86,20 +90,48 @@ export const signUpMachine = createMachine(
...
+              ),
+            ],
+          },
+          mnemonicNotMatch: {
+            entry: [
+              assignErrorMessage(
+                'The seed phrase not match check the words for typos or missing words'
```
LuizAsFight, line 105:
```suggestion
                'The seed phrase doesn't match. Check the phrase for typos or missing words'
```

LuizAsFight (changes requested):

LuizAsFight (approved):

PR #347, opened 2022-12-07
refactor: update the bn.format calls to use the default params

### LINKED ISSUE #198: Fix bn.format across the app
### Motivation

When [fuels-ts/595](https://github.com/FuelLabs/fuels-ts/issues/595) gets merged and released, some breaking changes will happen in some places that use `bn.format` function.

The reason is that now we should invert the way we use default parameters.

When using
```
bn.parseUnits("0.5").format({ precision: DECIMAL_UNITS });
```
Should replace to:
```
bn.parseUnits("0.5").format();
```

---------------

When using
```
bn.parseUnits("0.5").format();
```
Should replace to:
```
bn.parseUnits("0.5").format({ precision: 3, minPrecision: 1 });
```

tomiiide:
Fixes #198


luizstacio (changes requested):

LuizAsFight (changes requested):
code looks good.

recently @luizstacio introduced this function -> `formatFullAmount`, which:
- became pretty useless now with SDK changes
- it's being used only in 1 place

can you remove this function, and in the screen that uses it, replace to a simple `bn(amount).format()` ??

LuizAsFight (approved):

luizstacio (approved):

Issue #243, opened 2022-11-22
Check if wallet is correctly setup before accept new actions


luizstacio:
### What version of Wallet are you using?

0.0.1

### Steps to Reproduce

1. Install wallet. when the setup opens, close the page
2. Go to https://fuels-wallet.vercel.app/#/dapp and hit connect
3. It should open a popup empty and open the setup flow.
<img width="669" alt="Screenshot 2022-11-21 at 23 06 00" src="https://user-images.githubusercontent.com/3941923/203248101-06077f7c-648b-4d85-a9d3-3bf980ccefec.png">

### Expected Behavior

We could implement it in one of these ways;
1. When the connection action is triggered, but the wallet is not ready, throw an error: 'Wallet not ready.'
2. When the wallet is not ready the SDK is not injected on the window.

### Actual Behavior

After installing the wallet and not completing the setup flow, if the user tries to request a connection, it will show a broken popup and will open the setup flow.

PR #354, opened 2022-12-10
fix: check if wallet is setup before open connection

luizstacio:
closes: #243 

We have some issues with imports on services, if we import the root files on the services, it can't be used on the chrome extension once it didn't support front-end things. The main problem is the controlled field probably react-hook-form, that tries to access the window object.


packages/app/src/systems/Asset/utils/constants.tsx
```diff
@@ -1,31 +1,40 @@
...
+];
 
-export const ASSET_LIST = Object.values(ASSET_MAP);
+export const ASSET_MAP: Record<string, Asset> = {
+  '0x0000000000000000000000000000000000000000000000000000000000000000':
+    ASSET_LIST[0],
+  '0xe90f62d5216f58ffaf19e6eba07f944cc2cd36f9800b89419d52b6927f53976701':
+    ASSET_LIST[1],
+  '0x2260fac5e5542a773aa44fbcfedf7c193bc2c599': ASSET_LIST[2],
+};
```
LuizAsFight, line 40:
```suggestion
export const ASSET_MAP: Record<string, Asset> = {
  [ASSET_LIST[0].assetId]: ASSET_LIST[0],
  [ASSET_LIST[1].assetId]: ASSET_LIST[1],
  [ASSET_LIST[2].assetId]: ASSET_LIST[2],
};
```

LuizAsFight (changes requested):

LuizAsFight (approved):

PR #376, opened 2022-12-20
fix(app): store useSetMachineConfig

pedronauck:


packages/app/src/systems/CRX/background/services/BackgroundService.ts
```diff
@@ -93,7 +93,7 @@ export class BackgroundService {
     } else {
       await this.requireAccounts();
     }
-    return next(request, serverParams);
+    return next(request, serverParams!);
```
luizstacio, line 96: You can change the argument here: `serverParams: EventOrigin | undefined`. to `serverParams: EventOrigin`

luizstacio (changes requested):

luizstacio (approved):

PR #385, opened 2022-12-21
feat: improve balance visibility and hide account

luizstacio:
The current implementation of Balance visibility was using the isHiden prop inside the account specific. The balance visibility is a global setting, and because of that, it should be stored at a global level.

For this implementation, I had;
- Create utils Storage to wrap on localStorage with events enabling to be used with useSyncExternalStore
- Create a hook `useStorageItem` to wrap the use of a specific storage key value and setter.
- Using localStorage instead of indexedDB, in this case, looks better as it's a front-end-specific setting.
- Change accountMachine references names from setBalanceBivisibility to hideAccount.
- Create a new Component AmountVisibility to wrap value | '•••••'.
- Add AmountVisibility to Asset List amounts.


packages/app/src/systems/Account/components/BalanceWidget/BalanceWidget.tsx
```diff
@@ -90,22 +76,19 @@ export function BalanceWidget({
             </Text>
           </Copyable>
           <Text
-            color={isHidden ? 'gray10' : 'gray12'}
+            color={visibility ? 'gray12' : 'gray10'}
```
pedronauck, line 79: this style should be inside the `styles` object instead of on props

packages/app/src/systems/Account/components/BalanceWidget/BalanceWidget.tsx
```diff
@@ -90,22 +76,19 @@ export function BalanceWidget({
...
           >
-            <>
-              {account.balanceSymbol || '$'}&nbsp;
-              {isHidden ? '•••••' : formatAmount(account.balance)}
-            </>
+            {account.balanceSymbol || '$'}&nbsp;
+            <AmountVisibility value={account.balance} visibility={visibility} />
           </Text>
         </Flex>
         <Box css={{ marginRight: 6, marginTop: 8 }}>
```
pedronauck, line 87: the same here, styles should be inside `styles` obj

packages/app/src/systems/Core/hooks/useStorage.tsx
```diff
@@ -0,0 +1,15 @@
...
+import { useCallback, useSyncExternalStore } from 'react';
+
+import { Storage } from '../utils';
+
+export function useStorageItem<T = any>(key: string, defaultValue?: T) {
+  const state = useSyncExternalStore<T | null>(
+    Storage.subscribe,
+    () => Storage.getItem<T>(key) ?? defaultValue ?? null
+  );
+  const setState = useCallback((value: T) => {
```
pedronauck, line 11: why do you need a `useCallback` here?

packages/app/src/systems/Home/pages/Home/Home.tsx
```diff
@@ -21,16 +23,21 @@ export function Home() {
     navigate(Pages.receive());
   }, [navigate]);
 
+  const goToAccounts = useCallback(() => {
```
pedronauck, line 26: same here, i think you don't need a `useCallback` here... `useCallback` is better used when you have heavy computations using the component scope and want to avoid re-create these computations in all re-renders you can use it, but for simple cases like that you don't need... in fact, is very rare when you, in fact, will need it

pedronauck (changes requested):

packages/app/src/systems/Home/pages/Home/Home.tsx
luizstacio: I agree I used here just to keep consistency on the code.

pedronauck (approved):

PR #408, opened 2023-01-02
feat: create tx item component

### LINKED ISSUE #29: Create TransactionItem component
## Figma Reference
https://www.figma.com/file/uEckQY4EaQ5rXwyISg6d1g/Fuel-Wallet?node-id=77%3A6771

## Props
- `transaction: Transaction`
- `type: TransactionType`

tomiiide:
This PR creates the ActivityItem component.

Closes #29 

https://www.loom.com/share/6a8945b7843243e383818f5e9a38878d


luizstacio (changes requested):
- The loader component is slighter bigger than the normal component
- The icon should have colors like in the designs here. They are the same color, image bellow.
- I think we can remove the amount on the tx. This would cause some confusion as the tx could include multiple assets
- The label from/to should be in bold and have a space
- The address displayed should be in bech32 format. You can use the method from the class Address
- I think the font sizes are not matching the design in proportion. They look a bit bigger here.

Screenshot 2023-01-09 at 17 47 45

pedronauck (changes requested):
### Design considerations:

As you see comparing both the design we have on Figma and the result in your Storybook, there are a few things that are very different
- Padding bottom on your component is bigger than other paddings
- There's no difference in grays on texts (like we have on Figma)
- The font sizes are different from Figma also.
- The icon color is always gray. It should change according to status.
- There's no space between `To:<text>` in Storybook

| Figma | Storybook |
| -- | -- |
| CleanShot 2023-01-09 at 19 46 02@2x | CleanShot 2023-01-09 at 19 46 23@2x |

- The size isn't matching between the component and the loader, you can see if you change between stories quickly

Kapture 2023-01-09 at 19 47 19

packages/app/src/systems/Transaction/components/ActivityItem/ActivityItem.tsx
```diff
@@ -0,0 +1,140 @@
...
+
+  const formatDate = (date: Date) =>
+    `${date.toLocaleString('default', { month: 'short' })} ${date.getDate()}`;
+
+  const assetInfo = getAssetInfoById(amount?.assetId, amount);
+
+  return (
+    <Card css={styles.root} data-testid="activity-item">
+      <TxIcon operationName={mainOperation.name} />
+      <Flex direction="column" css={styles.contentWrapper}>
```
pedronauck, line 71: You can use the `<Stack>` component from `@fuel-ui` instead of `<Flex direction="column">` here

packages/app/src/systems/Transaction/utils/tx.ts
```diff
@@ -682,5 +691,32 @@ export function parseTx({
     isStatusPending: isStatusPending(gqlStatus),
     type,
     status,
+    time,
   };
 }
+
+export function getOperationDirection(
```
pedronauck, line 698: This function could be simplified into this:

```ts
export function getOperationDirection(
  operation: Operation,
  owner: string,
) {
  const ownerAddr = Address.fromString(ownerAddr)
  const operationAddr = operation?.to?.address || operation?.from?.address

  if (!owner?.length || !operationAddr) {
    return OperationDirection.unknown
  }

  return ownerAddr.equals(Address.fromString(operation.to.address))
    ? OperationDirection.to
    : OperationDirection.From
}
```

pedronauck (changes requested):

packages/app/src/systems/Transaction/utils/date.ts
```diff
@@ -0,0 +1,60 @@
+type DateDifference = {
```
luizstacio, line 1: Can we use some date lib to do this things?

packages/app/src/systems/Transaction/__mocks__/tx.ts
```diff
@@ -226,6 +230,61 @@ export const MOCK_TRANSACTION_CONTRACT_CALL: MockTransaction = {
...
+      {
+        ...MOCK_TRANSACTION_CONTRACT_CALL.tx.operations[0],
+        name: OperationName.receive,
+      },
+    ],
+    time: twoHoursAgo.toString(),
+    id: '0x18617ccc580478214175c4daba11903df93a66a94aada773e80411ed06b6adea',
+  },
+  receipts: MOCK_TRANSACTION_CONTRACT_CALL.receipts,
+};
```
luizstacio, line 279: You can create a function to abstract the creation of the tx and just pass the data you need to change will remove some of duplicated code here.

luizstacio (changes requested):

packages/app/src/systems/Transaction/utils/date.ts
tomiiide: alright, pulling in `day-js`

packages/app/src/systems/Transaction/__mocks__/tx.ts
tomiiide: hmm, thanks for this suggestion, I did that.

luizstacio (approved):

LuizAsFight:
- this is not a transaction ID, this is a address
- I see all examples in storybook showing `To` label.
  - when we receive a transaction, we receive FROM someone
  - when we send a transaction, we send TO someone

image

packages/app/src/systems/Transaction/utils/tx.ts
```diff
@@ -682,5 +691,20 @@ export function parseTx({
...
+  if (!owner?.length || !operationAddr) {
+    return OperationDirection.unknown;
+  }
+
+  const ownerAddr = Address.fromString(owner);
+
+  return ownerAddr.equals(Address.fromString(operation?.to?.address ?? ''))
+    ? OperationDirection.from
+    : OperationDirection.to;
+}
```
LuizAsFight, line 710: I think this function should live somewhere else than `tx.ts` file.

reasons:
- this code will be moved to SDK soon, and this function shouldn't be in SDK I think
- this function doesn't have unit tests (all the other methods in this file are extensively tested)

LuizAsFight (changes requested):

LuizAsFight (approved):

pedronauck (approved):

luizstacio (approved):

PR #409, opened 2023-01-02
feat: create activity list component

### LINKED ISSUE #359: Create TransactionList component
https://www.figma.com/file/uEckQY4EaQ5rXwyISg6d1g/Fuel-Wallet?node-id=290%3A15996&t=qxjn5NtpvjjTQNEm-4

Create a component that renders a list of TxItem components.

tomiiide:
This PR creates the ActivityList component, that shows the transactions list of an address.

closes #359 

https://www.loom.com/share/76b4f3069a9c45e09653ebe1a1ec5aea


packages/app/src/systems/Transaction/components/ActivityList/ActivityList.stories.tsx
```diff
@@ -0,0 +1,35 @@
+import type { Story } from '@storybook/react';
+
+import { MOCK_TXS } from '../../__mocks__/transactions';
+
+import type { ActivityListProps } from './ActivityList';
+import { ActivityList } from './ActivityList';
+
+import { MOCK_ACCOUNTS } from '~/systems/Account';
+
+export default {
```
pedronauck, line 10: You can write this story using a `Template` component like Storybook recommend here:

https://storybook.js.org/docs/react/writing-stories/introduction
CleanShot 2023-01-09 at 20 11 42@2x

And add a `<Box css={{ maxWidth: 300 }}>` to don't have the component in a full-screen window.

packages/app/src/systems/Transaction/components/ActivityList/ActivityList.test.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+  it('should render the same amount of activity items as transactions', async () => {
+    render(
+      <ActivityList
+        transactions={MOCK_TXS}
+        isLoading={false}
+        ownerAddress={MOCK_ACCOUNTS[3].address}
+      />
+    );
+    expect(screen.getAllByTestId('activity-item')).toHaveLength(
```
pedronauck, line 67: You should always use `aria-label` instead of `data-testid` in order to follow `testing-library` guidelines:
https://testing-library.com/docs/queries/about#priority

packages/app/src/systems/Transaction/components/ActivityList/ActivityList.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+export const ActivityList = ({
+  transactions,
+  isDevnet = false,
+  isLoading,
+  ownerAddress,
+}: ActivityListProps) => {
+  if (isLoading) return <ActivityList.Loading />;
+
+  const isEmpty = !transactions || !transactions.length;
```
pedronauck, line 25:
```suggestion
  const isEmpty = !transactions?.length;
```

packages/app/src/systems/Transaction/components/ActivityList/ActivityList.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+  if (isEmpty) return <ActivityList.Empty isDevnet={isDevnet} />;
+
+  return (
+    <Flex gap={'$2'} direction={'column'}>
```
pedronauck, line 30:
```suggestion
    <Stack gap="$2">
```

packages/app/src/systems/Transaction/utils/icon.ts
```diff
@@ -0,0 +1,19 @@
+import { Icon } from '@fuel-ui/react';
+
+import { OperationName } from './tx.types';
+
+export const getTxIcon = (type?: OperationName) => {
```
pedronauck, line 5: You could use an object mapping instead a switch here:

```tsx
const ICON_MAP = {
  [OperationName.transfer]: 'UploadSimple',
  [OperationName.receive]: 'DownloadSimple',
  [OperationName.mint]: 'ArrowRight',
  [OperationName.predicatecall]: 'ArrowsLeftRight',
  [OperationName.contractCall]: 'UploadSimple'
}

export const getTxIcon = (type?: OperationName) => {
  return ICON_MAP[type] ?? 'ArrowRight'
}
```

pedronauck (changes requested):

packages/app/src/systems/Transaction/components/ActivityList/ActivityListEmpty.tsx
```diff
@@ -0,0 +1,67 @@
+import { cssObj } from '@fuel-ui/css';
```
pedronauck, line 1: I guess the illustration size is slighter different than Figma

CleanShot 2023-01-09 at 20 10 29@2x

packages/app/src/systems/Transaction/components/TxIcon/TxIconLoader.tsx
```diff
@@ -0,0 +1,17 @@
+import { darkColors } from '@fuel-ui/css';
+import type { IContentLoaderProps } from 'react-content-loader';
+import ContentLoader from 'react-content-loader';
+
+export const TxIconLoader = (props: IContentLoaderProps) => (
```
pedronauck, line 5: I didn't understand why you need a Loader for `TxIcon` component? 🤔 Because since you already have a loader component for the `ActivityItem` I guess you don't need another one here.

pedronauck (changes requested):

packages/app/src/systems/Transaction/components/ActivityList/ActivityList.test.tsx
tomiiide: I tried, but I `a11y` tests were failing as you can't put an `aria-label` on a `div`.

tomiiide: would remove this test case, as there's no determinate way to write the test without `test-id`s.

tomiiide: I ended up just counting the number of children the ActivityList element has as a test.

packages/app/src/systems/Transaction/components/ActivityList/ActivityListEmpty.tsx
tomiiide: Screenshot 2023-01-11 at 5 37 33 PM


Made it the right size.

packages/app/src/systems/Transaction/components/ActivityList/ActivityList.tsx
```diff
@@ -0,0 +1,43 @@
...
+export interface ActivityListProps {
+  transactions: Tx[];
+  providerUrl?: string;
+  isLoading?: boolean;
+  isDevnet?: boolean;
+  ownerAddress: string;
+}
+
+export const ActivityList = ({
+  transactions,
```
LuizAsFight, line 18: can we rename to `txs` ?

packages/app/src/systems/Transaction/utils/icon.ts
```diff
@@ -1,19 +1,15 @@
...
+const ICON_MAP = {
+  [OperationName.transfer]: 'UploadSimple',
+  [OperationName.receive]: 'DownloadSimple',
+  [OperationName.mint]: 'ArrowRight',
+  [OperationName.predicatecall]: 'MagicWand',
+  [OperationName.contractCall]: 'ArrowsLeftRight',
+};
+
+export const getTxIcon = (type?: OperationName): Icons => {
+  return type ? ICON_MAP[type] ?? 'ArrowRight' : 'ArrowRight';
```
LuizAsFight, line 14: wouldn't this code have the same effect?

`return ICON_MAP[type] || 'ArrowRight'`

LuizAsFight (changes requested):

LuizAsFight:
1- image looks blurred, like if it's rendered in wrong ratio? can we export the image directly in the correct size and use it without changing size?
2- text mentions bridge, but we don't actually have one yet @luizstacio ? I think for text we should keep only "You don't have any activity yet"

image

luizstacio:
> 

You can remove the link to the bridge for now.

packages/app/src/systems/Transaction/utils/icon.ts
tomiiide: It would but typescript requires a check as  `type` can be `undefined`, and there's not `undefined` key in `ICON_MAP`.

Screenshot 2023-01-17 at 3 39 37 AM

LuizAsFight (dismissed):

luizstacio (approved):

pedronauck (approved):

luizstacio (approved):

PR #414, opened 2023-01-03
feat: add getSelectedAccount method to Wallet SDK

### LINKED ISSUE #412: Add function on fuel instance to get selected account
### Motivation

Dapp frontends need to know which account is currently selected in order to send transactions/sign messages from the correct account.  Currently there is no way to do this.

### Usage example

```
const currentAccount = await window.fuel.currentAccount();
```

### Possible implementations

I'm not really sure how this would be implemented.  The internal app db has this info, but I'm not sure how to expose that command to the wallet sdk.

matt-user:
closes #412


.changeset/late-planets-rescue.md
```diff
@@ -0,0 +1,2 @@
+---
```
luizstacio, line 1: Add a description to this change as it touches on the SDK and adds a new method.

It could be: `Add to get selected account from wallet via SDK`

packages/app/src/systems/CRX/background/services/BackgroundService.ts
```diff
@@ -189,4 +190,9 @@ export class BackgroundService {
     );
     return signedMessage;
   }
+
+  async getSelectedAccount() {
+    const selectedAccount = await AccountService.getSelectedAccount();
+    return selectedAccount;
```
luizstacio, line 196: We should only return the address of the account.

luizstacio (changes requested):

packages/docs/docs/api.mdx
```diff
@@ -42,6 +42,13 @@ sign, and send the transaction.
 
 <CodeImport file="../../sdk/src/Fuel.test.ts" testCase="sendTransaction" />
 
+### Get Selected Account
+
+`async getSelectedAccount(): Promise<Account>`<br />
```
luizstacio, line 47:
```suggestion
`async getSelectedAccount(): Promise<string>`<br />
```

luizstacio (changes requested):

luizstacio (approved):

PR #423, opened 2023-01-09
feat: update docs

luizstacio:
Update docs to follow patterns. Change the order of sections and remove them from the how-to.


packages/docs/docs/how-to-use.mdx
```diff
@@ -48,9 +48,3 @@ To transfer an amount to other address, use the `wallet.transfer` method and pas
 <CodeImport file="../examples/Transfer.tsx" lineStart="31" lineEnd="39" />
 
 <Examples.Transfer />
-
-<CodeImport
```
matt-user, line 52: Shouldn't we still include getSelectedAccount in the how to use section of the docs

matt-user (approved):

Issue #393, opened 2022-12-22
Currently when user changes network DApp application can't use the new network


luizstacio:
The current connect method allows devs to inform a providerUrl but it didn't take effect and causes an issue when trying to assign the providerUrl.

We should make the providerUrl been sync with the Wallet provider from inside the Wallet and enforce devs to request network selection.

Issue #391, opened 2022-12-22
In connect screen, it breaks when click on network button


LuizAsFight:
### What version of Wallet are you using?

0.0.1

### What is happening?

When click in `Localhost` network button
<img width="334" alt="image" src="https://user-images.githubusercontent.com/8636507/209027128-87ad3960-0937-4af2-be4f-a6e615a65fd1.png">

popup closes and show following error in doc page
![image](https://user-images.githubusercontent.com/8636507/209027157-fee3abf2-fdd9-43a7-8a15-d0937a791207.png)

PR #424, opened 2023-01-09
feat: sync DApp provider with wallet provider

luizstacio:
closes: #393 #391


LuizAsFight:
I'm testing in latest PR preview docs/extension

context: I added a new network "Localhost" pointing 

problems in connect screen:
- when click in "Localhost", nothing happens
- when click in "account" nothing happends
- after connecting, when trying to disconnect, an error is shown


https://user-images.githubusercontent.com/8636507/212160210-bd159fe6-cabf-4983-97bb-930a7918348c.mov

LuizAsFight:
in same context previously mentioned:

when click in "transfer asset", a error shows up "Network request Failed".
Debugging it I see "ERR_BLOCKED_BY_CLIENT" error, probably configs between domains, some block from chrome because it's a localhost server... just wanna make sure this is expected or if we should fix

luizstacio:
> I'm testing in latest PR preview docs/extension
> 
> context: I added a new network "Localhost" pointing
> 
> problems in connect screen:
> 
> * when click in "Localhost", nothing happens
> * when click in "account" nothing happends
> * after connecting, when trying to disconnect, an error is shown
> 
>  Screen.Recording.2023-01-12.at.16.14.38.mov

- The network selector was disabled, which is not supposed to work, we should remove the cursor pointer also, and will do it.
- The error about the network being blocked is on the browser level you can enable it by clicking on the Brave shield on the side of the URL.
- The accounts button is not related to this PR but I open an issue for it: https://github.com/FuelLabs/fuels-wallet/issues/432

LuizAsFight (changes requested):
code looks ok but I asked functional changes in prev comments... pls ask review again when finished @luizstacio

packages/sdk/src/Fuel.ts
```diff
@@ -1,16 +1,40 @@
...
+      return FuelWeb3Privates.provider;
+    }
+    // Otherwise, create a new provider instance
+    // fetch the current network and connect the provider
+    const provider = new FuelWalletProvider(providerConfig.url, this);
+    FuelWeb3Privates.provider = provider;
+
+    // Listen for network changes and connect the provider
+    // selected network from the user
+    this.on('network', async (network) => {
```
matt-user, line 28: Would it be possible for the network to change without a call to `getProvider`?  Basically I'm asking would it be possible for a network change event to fire before reaching this code which handles network change events.  Sorry if this is a dumb question lol.

luizstacio: As the getProvider always returns the same instance when the event is triggered and the URL has changed, the changes will have an instant effect.

If the event is triggered before riching this point would not be a problem as the current network is requested from the wallet at the time we create the instance of the provider line 16.

matt-user (approved):
just tested locally and it works 😄

LuizAsFight (approved):
lgtm

PR #425, opened 2023-01-09
chore: remove storybook on docs production

luizstacio:
The main focus of the live docs is just the wallet usage and installation. The storybook is not related to it. Because of this, we should hide it on the main website.


tomiiide (approved):
LGTM!

Issue #244, opened 2022-11-22
Update unlock flow, to request password once per session


luizstacio:
Currently, the wallet asks for users a password every time the user is required to sign a transaction or message.

We should move it to a session and manage the global state; once the user unlocks the wallet, we start a session that auto-closes after a configured period. Default (~5min).

luizstacio:
We need to make sure the state is live only on memory. This will require moving the access of the signature to the service worker.

pedronauck:
I'm working in a solution for this already

PR #426, opened 2023-01-10
refactor(app): use unlock as a global machine

pedronauck:
### Main changes

- Unlock now is a global machine and also `UnlockDialog`. Now we don't have more `unlockMachine` as submachines for each time we need to unlock, just need to call `store.unlock()` in order to open unlock popup.
- Now user just needs to unlock one time, if the wallet is already unlocked, unlock is bypassed.
- There's a new `UnlockService` that gets the manager and wallet directly from the store.

### Example of usage

```tsx
function addAccount(input: AddAccountInput) {
  store.unlock({
    onSuccess() {
      service.send('ADD_ACCOUNT', { input })
    }
  }
}

const manager = await UnlockService.getManagerUnlocked();
const wallet = await UnlockService.getWalletUnlocked();
```


luizstacio:
I'm working on a solution to keep the wallet unlocked on the extension, as this PR will not keep the wallet unlocked between sessions. We can hold to implement this in an integrated way with the unlock between sessions.

luizstacio:
Going to close this PR as it is being addressed on the new flow: #244

PR #456, opened 2023-01-20
fix: make unlockDialog fullscreen only when the browser is not fullscreen

### LINKED ISSUE #199: macOS approve screen not center-aligned when fullscreen

0.0.1

### Steps to Reproduce

1- ask to connect with the extension
2- you will see this:

image

### Expected Behavior

connect screen should be centered

### Actual Behavior

connect screen is not centered

### LINKED ISSUE #201: macOS unlock screen too big when in fullscreen

0.0.1

### Steps to Reproduce

Unlock screen is too big when on fullscreen

image

### Expected Behavior

fullscreen should respect app box size

### Actual Behavior

it's not respecting app box size

LuizAsFight:
- make unlockDialog not fullscreen when browser is in fullscreen mode
- centralize app in `body` element: `margin: 0 auto`;

Closes #201 
Closes #199

### DIFF
--- a/packages/app/src/systems/Core/components/UnlockDialog/UnlockDialog.tsx
+++ b/packages/app/src/systems/Core/components/UnlockDialog/UnlockDialog.tsx
@@ -120,6 +120,7 @@ const styles = {
           }
         : {
             width: '350px',
+            height: '600px',
           }),
     });
   },
--- a/packages/app/src/systems/Core/components/UnlockDialog/UnlockDialog.tsx
+++ b/packages/app/src/systems/Core/components/UnlockDialog/UnlockDialog.tsx
@@ -14,6 +14,8 @@ import { useUnlockForm } from '../../hooks';
 import type { UnlockFormValues } from '../../hooks';
 import { UnlockForm } from '../UnlockForm';
 
+import { IS_CRX_POPUP } from '~/config';
+
 export type UnlockDialogProps = {
   title?: string;
   unlockText?: string;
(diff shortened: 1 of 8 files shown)


packages/app/src/systems/Core/components/UnlockDialog/UnlockDialog.tsx
```diff
@@ -116,7 +120,6 @@ const styles = {
           }
         : {
             width: '350px',
-            height: '600px',
```
matt-user, line 119: Why was the height removed, but not the width?

matt-user (commented):
Does this pr need a changeset?

packages/app/src/systems/Core/components/UnlockDialog/UnlockDialog.tsx
LuizAsFight: because 600px was put initially to match extension popup size, but later we made the fullscreen mode to cover it.

here is with 600px fixed
image

here with no height fixed:
image

we don't need to set fixed height as:
- height will be automatic calculated by content
- we don't need extra space

LuizAsFight: @matt-user FYI I rollback this height change... I'm keeping 600px as it was looking bad after centering app

LuizAsFight:
@matt-user we're not doing changesets to `app` package yet

matt-user (approved):

luizstacio (changes requested):
Testing using the send transaction flow shows the unlock as fullscreen yet. I think the correct way to fix this issue is by making the unlock dialog respect the layout container and render only on the application area.
Screenshot 2023-01-22 at 20 40 41
Screenshot 2023-01-22 at 20 42 38

luizstacio (approved):

PR #473, opened 2023-01-23
feat(sdk): add `currentAccount` function/event to sdk

### LINKED ISSUE #419: Add way to monitor when the selected account changes
### Motivation

Frontend devs using the sdk need to know when the selected account changes in order to sign/send transactions from the correct wallet/account.

### Usage example

I'm not sure exactly how usage of this feature would look, but something like
```
fuel.on('selectAccount', () => {
    updateWallet();
});
```

### Possible implementations

Another type would have to be added to `FuelEvents` [here](https://github.com/FuelLabs/fuels-wallet/blob/master/packages/types/src/fuel.ts#L8) and then the event would have to be emitted somewhere.

LuizAsFight:
This PR introduces new function to sdk:
`currentAccount` - returns the current account being used in the extension

also introduces new event `currentAccount`:

```
fuel?.on('currentAccount', (currentAccount: string) => {
  // your code here
});
```

This PR also change names from `selectedAccount` to `currentAccount`, in order to match a more realistic naming

Closes #419


packages/app/src/systems/Account/machines/accountMachine.tsx
```diff
@@ -88,7 +88,7 @@ export const accountMachine = createMachine(
             actions: ['hideAccount'],
             target: 'idle',
           },
-          SELECT_ACCOUNT: {
+          SET_CURRENT_ACCOUNT: {
             target: 'selectingAccount',
```
matt-user, line 92:
```suggestion
            target: 'settingAccount',
```
or maybe
```suggestion
            target: 'settingCurrentAccount',
```

packages/app/src/systems/Account/machines/accountMachine.tsx
```diff
@@ -148,7 +148,7 @@ export const accountMachine = createMachine(
       },
       selectingAccount: {
```
matt-user, line 149: Change to one of the suggestions above

packages/app/src/systems/CRX/background/services/DatabaseEvents.ts
```diff
@@ -51,5 +52,35 @@ export class DatabaseEvents {
         ])
       );
     });
+
+    this.databaseObservable.on('accounts:update', async (updateEvent) => {
+      // Broadcast only if it's the current account
+      if (!updateEvent.obj.isCurrent) return;
+
+      const [currentAccount, connections] = await Promise.all([
+        AccountService.getCurrentAccount(),
```
luizstacio, line 61: You can get the data directly from the update event.obj removing the need to query the data again.

packages/app/src/systems/CRX/background/services/DatabaseEvents.ts
```diff
@@ -51,5 +52,35 @@ export class DatabaseEvents {
...
+        ConnectionService.getConnections(),
+      ]);
+      const origins = connections
+        .filter(
+          (connection) =>
+            connection.accounts.indexOf(currentAccount?.address || '') !== -1
```
luizstacio, line 67:
```suggestion
            connection.accounts.includes(currentAccount?.address || '')
```

luizstacio (changes requested):
Can we add a e2e test on `crx.test.ts` to see if the event triggers?

packages/app/src/systems/CRX/background/services/DatabaseEvents.ts
LuizAsFight: _that makes sense_

luizstacio (approved):

PR #505, opened 2023-01-24
feat: update accounts after send transactions

### LINKED ISSUE #439: After sending a transaction using the wallet send feature, the balance is not updated
After sending a transaction and going back to the home page, the user balance should be re-loaded, to reflect the new balance.

tomiiide:
This PR updates the account balance after navigating from completing a send transaction.

closes #439

### DIFF
--- a/packages/app/src/systems/Send/components/SendConfirm/SendConfirm.tsx
+++ b/packages/app/src/systems/Send/components/SendConfirm/SendConfirm.tsx
@@ -1,7 +1,6 @@
 import { Button, Text } from '@fuel-ui/react';
 import { useNavigate } from 'react-router-dom';
 
-import { store } from '~/store';
 import { ContentHeader, Layout, Pages } from '~/systems/Core';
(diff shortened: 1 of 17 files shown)


packages/app/src/systems/Send/components/SendConfirm/SendConfirm.tsx
```diff
@@ -13,6 +14,12 @@ export function SendConfirm({ txRequest }: SendConfirmProps) {
   const amountSent = txRequest.ethAmountSent;
   const isDone = txRequest.status('success') || txRequest.status('failed');
   const navigate = useNavigate();
+
+  const goToHome = () => {
+    store.updateAccounts();
```
luizstacio, line 19: This should be done inside the txRequest machine right after success or failure step.

luizstacio (changes requested):

luizstacio (approved):

PR #513, opened 2023-01-27
chore: improve project tooling

pedronauck:
## Tasks

This pull request makes some changes to improve precise details on our tooling

### `.lintstagedrc`
- [x] Change to `prettier --write` and `eslint --fix`. So, instead of just checking on husky, it will format and fix the files before committing.
- [x] Add other formats on `lintstaged` to run `prettier`

### `.prettierrc`
- [x] Setup prettier to format all types of files, since we have a bunch new of them and this configuration  came from `fuels-ts` that's a totally different type of project

### `.vscode` folder
- [x] Add a recommendation of a new extension that is a [Jest runner](https://marketplace.visualstudio.com/items?itemName=firsttris.vscode-jest-runner) inside VSCode. I used this extension on `fuel-ui` and it's capable to debug using VSCode and Jest, it's very useful for debugging without console.log on tests, and this extension is way better than the with most downloads on VSCode extensions store.


.vscode/settings.json
```diff
@@ -1,7 +1,11 @@
 {
   "editor.defaultFormatter": "esbenp.prettier-vscode",
   "prettier.prettierPath": "./node_modules/prettier",
+  "editor.formatOnPaste": false,
+  "editor.formatOnType": false,
   "editor.formatOnSave": true,
+  "editor.formatOnSaveMode": "file",
+  "files.autoSave": "onFocusChange",
```
LuizAsFight, line 8: I think we don't need this line... saving should be intentionally fired manually

what you think @luizstacio  ??

I use cmd+shift+s when I want to save all files

LuizAsFight (changes requested):
minor comment

.vscode/settings.json
pedronauck: Hm, but in which way this can change the `ctrl+z/ctrl+shift+z`? I didn't get it 🤔

LuizAsFight: it helps identifying what files you changed last, then use undo/redo if needed...
if applying this change everything will be always saved when change file right ?

LuizAsFight: also many times we need to change a lot of files to get a change done... if autoSave is enabled, live reload will break app you're running a lot of times...

saving should be manual, if want can execute `cmd+shift+s` to save all files, that's what I do when complete changes in multiple files

pedronauck: I always used this personal setting on my VSCode and I never have problems with this. Some extensions recommend this type of autosave:
CleanShot 2023-01-26 at 22 58 04

This is good to ensure that, if something VSCode you don't lose what you did I guess also.
About the live reloading, I don't think this is a point.

But idk, is a point of view, I can remove it, no problem.

pedronauck: Let's see whats the rest think about @tomiiide @luizstacio

LuizAsFight: we can always add later if @FuelLabs/frontend think it's good.. I personally don't like it

can wait for other to answer 

or just remove for now and add later in other PR, then we can quickly approve this one

LuizAsFight: thing with `ctrl+z / ctrl+shift+z` is that if the file is unsaved, you know how many `undo` you need to get to the previous saved(valid) one

LuizAsFight: ask for re-review when this is solved

LuizAsFight: sounds good @pedronauck

luizstacio: I also don't like auto-save, we probably should remove it from the project itself.

.changeset/silent-sloths-flow.md
```diff
@@ -1,6 +1,6 @@
 ---
-"@fuel-wallet/types": minor
-"@fuel-wallet/sdk": minor
+'@fuel-wallet/types': minor
+'@fuel-wallet/sdk': minor
```
luizstacio, line 3: This will not be an issue for the changesets?

pedronauck: why?

.vscode/settings.json
pedronauck: removed

.changeset/silent-sloths-flow.md
luizstacio: Is more a question as I don't know how changeset cli parsers

luizstacio (dismissed):

.vscode/settings.json
```diff
@@ -1,7 +1,6 @@
 {
   "editor.defaultFormatter": "esbenp.prettier-vscode",
   "prettier.prettierPath": "./node_modules/prettier",
-  "editor.formatOnSave": true,
```
LuizAsFight, line 4: it's good to keep `formatOnSave` ,

my prev comment was about `autoSave` instead

LuizAsFight (changes requested):

LuizAsFight (approved):
great work!

PR #514, opened 2023-01-27
chore: update deps

pedronauck:


packages/app/src/systems/Core/machines/fetchMachine.test.ts
```diff
@@ -50,7 +50,7 @@ describe('fetchMachine', () => {
     expect(fetchSpy).toHaveBeenCalledTimes(3);
   });
 
-  it('should fail after attempting 5 times', async () => {
+  it.only('should fail after attempting 5 times', async () => {
```
LuizAsFight, line 53:
```suggestion
  it('should fail after attempting 5 times', async () => {
```

LuizAsFight (changes requested):

packages/app/src/systems/Core/machines/fetchMachine.test.ts
LuizAsFight: would be nice to remove orthogonal changes to dependency updates, keeping this PR cleaner.

those changes can be done in other PR.
I see some breaking tests, maybe this changes are the reason ?

pedronauck: This was failing after update deps, because of that I needed to fixed,

packages/docs/src/lib/code-import.ts
```diff
@@ -100,11 +99,11 @@ export function codeImport(options: Options = { filepath: '' }) {
...
 
   return function transformer(tree: Root) {
-    const nodes: [any, number | null, Parent][] = [];
+    const nodes: [any, number | null, any][] = [];
 
     visit(tree, 'mdxJsxFlowElement', (node: any, idx, parent) => {
       if (node.name === 'CodeImport') {
-        nodes.push([node as any, idx, parent as Parent]);
+        nodes.push([node as any, idx, parent as any]);
       }
```
tomiiide, line 107: why are we using the `any` type?

pedronauck: because was removed from the package 😕

tomiiide (approved):

LuizAsFight (approved):

Issue #452, opened 2023-01-19
Wallet SDK should allow only connected accounts and selected accounts to execute actions


luizstacio:

luizstacio:
close in favor of #495

Issue #417, opened 2023-01-04
Secure connections just to the address in the allowed list


luizstacio:
When a Dapp requests permission to access the Wallet, the user permits to select addresses. Still, today, the DApp can request transactions, signature messages, or get the active wallet that is not allowed in the connection.

These rules should be handled on the background service, ensuring the DApp is sandboxed to the allowed addresses.

Issue #495, opened 2023-01-24
SDK scoped address features


luizstacio:
When calling `signMessage` or `sendTransaction` the SDK should inform, which address/account the application is requesting this action.

On the Application side, this information should be;
- Validate if the connection (dapp) has access to this account, if not it should reject the request.
- Use this address/account to make the action request.

PR #517, opened 2023-01-28
feat: add checks for address permissions on sendTransaction and signMessage request

luizstacio:
- Implemented validation for `sendTransaction` and `signMessage` on the Wallet level to ensure authorized account access only.
- Enhanced CRX e2e tests for improved handling of desired and rejected flows.

Note: This work serves as the foundation for addressing #495, which will enable proper functionality for address-based methods.

close: #417
close: #452


packages/sdk/src/utils/getTransactionSigner.ts
```diff
@@ -0,0 +1,21 @@
...
+      default:
+        return undefined;
+    }
+  })[0];
+
+  if (!address) {
+    throw new Error('No possible signer found!');
+  }
+
+  return Address.fromB256(hexlify(address)).toString();
```
LuizAsFight, line 20: this will always use [0] signer, is there a multi-signature level transaction that we need to cover? if yes can we create a follow up issue on that ?

LuizAsFight: also this code could be inside `tx utils` , maybe adding some specific methods to deal with `TransactionRequest` / `TransactionRequestLike`

LuizAsFight (changes requested):
change request but can be done in other issue

packages/sdk/src/utils/getTransactionSigner.ts
luizstacio: I will not move to tx utils right now as it would make it difficult to use on the SDK creating a cycle dependency between SDK and Wallet once it's moved to `fuels-ts` we can replace this.

Open a issue for cover the multi-sig on the future https://github.com/FuelLabs/fuels-wallet/issues/522

LuizAsFight (approved):

PR #524, opened 2023-01-31
fix: menu click breaking app during load

### LINKED ISSUE #430: App break when click on menu before loading completes

0.0.3

### What is happening?

When clicking on `Menu` button before loading complete, it breaks

### steps to reproduce

https://user-images.githubusercontent.com/8636507/212939290-095061a0-b6ce-4b64-afe1-c1c6942569ab.mov

### How it should behave?

should just open the menu, or hide if it shouldn't be opened while loading.

Mac/Apple Desktop

### What browser?

Brave

_No response_

tomiiide:
Fixes  #430 

App was breaking because the `Avatar.Generated` component requires an `hash` of the account address.

The `account` is undefined, and we are trying to access the `account!.address` property while the account is loading.


LuizAsFight (approved):
lgtm

PR #528, opened 2023-02-01
feat: update unlock flow

luizstacio:
The main feature implemented on this PR is the change of the unlock flow to a single unlock every 15 minutes. close: #244 

On this PR;
- Remove unlock flows inside machines, as the vault should always be unlocked.
- Create a reset wallet (forget the password) screen.
- Add a lock button on the sidebar
- Split account machine and signup action to avoid dependency on VaultService inside accounts and have better code split.
- Remove all actions related to Vault from the AccountService
- Fix page flicker at the opening stage by adding a background and fixed sizes to the popup. close: #553 
- Fix all tests related to unlocking features
- Move request machines to a global store. This is needed to start the action before the wallet is unlocked
- In the development web `http://localhost:3000`, add a mock to keep the wallet unlocked all the time, not requiring passwords.


packages/app/src/systems/Network/services/network.ts
```diff
@@ -53,7 +53,7 @@ export class NetworkService {
         ...(count === 0 && { isSelected: true }),
         id: uniqueId(),
       });
-      return db.networks.get(id);
+      return db.networks.get(id) as Promise<Network>;
```
matt-user, line 56: Why do we type cast here with `as Promise<Network>` as well as in other places?

packages/app/src/systems/Settings/machines/setingsMachine.test.tsx
```diff
@@ -1,11 +1,10 @@
 import type { InterpreterFrom } from 'xstate';
```
matt-user, line 1: this filename has a typo, it should be names `settingsMachine.test.tsx`

packages/app/src/systems/Settings/machines/setingsMachine.test.tsx
```diff
@@ -23,68 +22,43 @@ const WORDS = [
...
-    );
-
-    const { event } = service.getSnapshot();
-
-    expect(event.data?.length).toBeGreaterThan(1);
-    expect(matches('done')).toBeTruthy();
+    expect(state.context.words.length).toBeGreaterThan(10);
   });
 
   it('should change the password of the user and we should be able to unlock it agan', async () => {
```
matt-user, line 54:
```suggestion
  it('should change the password of the user', async () => {
```
The wallet should already be unlocked after we change the password so I think the test name can be changed.

packages/app/src/systems/Settings/machines/settingsMachine.ts
```diff
@@ -34,6 +32,9 @@ export const settingsMachine = createMachine(
     predictableActionArguments: true,
     id: 'settingsMachine',
     initial: 'idle',
+    context: {
+      words: [],
```
matt-user, line 36: do we store the mnemonic in the settingsMachine context?  How safe is this?

packages/app/src/systems/Settings/machines/settingsMachine.ts
```diff
@@ -98,41 +99,28 @@ export const settingsMachine = createMachine(
...
           }
-          const secret = await AccountService.exportVault(input);
-
-          if (!secret) {
-            throw new Error('No Secret Found');
-          }
-
+          const secret = await VaultService.exportVault({
+            ...input,
+            vaultId: 0,
```
matt-user, line 122:
```suggestion
            vaultId: 0, // TODO change once we add multiple vault management
```

packages/app/src/systems/Settings/pages/ChangePassword/ChangePassword.tsx
```diff
@@ -81,11 +81,12 @@ export function ChangePassword() {
             </Alert>
             <ControlledField
               control={control}
-              name="oldPassword"
+              name="currentPassword"
               label="Current Password"
               render={({ field }) => (
                 <InputPassword
                   {...field}
+                  autoComplete="current-password"
```
matt-user, line 89: interesting... does this grab a users saved password from their browser?

packages/store/src/StoreClass.tsx
```diff
@@ -253,6 +255,63 @@ export class StoreClass<T extends MachinesObj> implements IStore<T> {
...
+        failureMessage = 'error',
+        timeout = 5000,
+      } = {},
+    ]: WaitForStateArgs<T, K>
+  ) {
+    try {
+      const service = this.services[key];
+      const doneState = done || 'done';
+      const failureState = failure || 'failed';
+      const failureMessageField = failureMessage || 'error';
```
matt-user, line 287:
```suggestion
      const doneState = done';
      const failureState = failure;
      const failureMessageField = failureMessage;
```

Can we do this bc of the default assignments in the function definition?

packages/store/src/StoreClass.tsx
```diff
@@ -253,6 +255,63 @@ export class StoreClass<T extends MachinesObj> implements IStore<T> {
...
+          cause: 'CustomState',
+        });
+      }
+
+      return appState as S;
+    } catch (err: unknown) {
+      if (err && (err as { cause: string }).cause === 'CustomState') {
+        throw err as Error;
+      }
+      throw new Error(
```
matt-user, line 310: Is there a way to tell that the inactivity error is actually the error we are encountering?  I'm just worried that this will throw the inactivity error when it is actually something else failing.

packages/store/src/types.ts
```diff
@@ -110,3 +110,13 @@ export type WaitForArgs<T extends MachinesObj, K extends keyof T> = [
   givenState: (state: StateFrom<T[K]>) => boolean,
   timeout?: number
 ];
+
+export type WaitForStateArgs<T extends MachinesObj, K extends keyof T> = [
+  key: K,
+  conifg?: {
```
matt-user, line 116:
```suggestion
  config?: {
```

LuizAsFight:
1- x button is not centralized
2- bold action phrase could be a separated paragraph, or even better could use fuel-ui Alert with warning status.

image

packages/app/playwright/e2e/SelectWallet.test.ts
```diff
@@ -34,7 +34,12 @@ test.describe('SelectWallet', () => {
     await getByAriaLabel(page, 'Accounts').click();
     await hasText(page, data.accounts[0].name);
     await hasText(page, data.accounts[1].name);
-    await getByAriaLabel(page, data.accounts[1].name).click();
+    await getByAriaLabel(page, data.accounts[1].name).click({
+      position: {
```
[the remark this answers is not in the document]
LuizAsFight: because it has a copy button in the account element, sometimes playwright will click on it and test will break.

packages/app/src/systems/Settings/pages/ChangePassword/ChangePassword.tsx
luizstacio: yes this helps password managers to auto complete

packages/store/src/StoreClass.tsx
luizstacio: This is a good idea, I will take a look at doing this.

packages/app/src/systems/Network/services/network.ts
luizstacio: This is because the return can be undefined by default, but on this cases it never returns undefined as it was just add.

packages/app/src/systems/Settings/machines/settingsMachine.ts
luizstacio: GQ! This is only on memory during the time you are seeing it. After you close the popup, the context is reset.

packages/app/src/systems/Vault/connectors/VaultWebConnector.ts
```diff
@@ -0,0 +1,31 @@
...
+  vault: VaultServer;
+  readonly clientVault: VaultClient;
+
+  constructor(clientVault: VaultClient) {
+    this.clientVault = clientVault;
+    this.vault = new VaultServer();
+    this.clientVault.onRequest = this.onRequest.bind(this);
+
+    // Mock unlock in development to save password
+    // on sesstion Storage
```
LuizAsFight, line 19:
```suggestion
    // on session Storage
```

packages/store/src/StoreClass.tsx
LuizAsFight: I think we should keep because default assignments in function definition will work unless you define something.... so for example: `done: ''` would bypass it

packages/store/src/StoreClass.tsx
LuizAsFight: if it's something else that we know, would be detected in first conditional, right @luizstacio ??

LuizAsFight (changes requested):

packages/app/src/systems/Unlock/components/ResetDialog/ResetDialog.tsx
```diff
@@ -0,0 +1,83 @@
...
+              by using your seed phrase.{' '}
+              <b>
+                This action will remove all data stored on this device,
+                including your seed phrase, accounts, networks and other
+                settings.
+              </b>
+            </Text>
+            <Text>
+              Make sure you had backed up your seed phrase before proceeding.
+            </Text>
```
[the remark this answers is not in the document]
matt-user: Also it should be "Make sure you have backed up your seed phrase before proceeding." instead of "had"

packages/store/src/StoreClass.tsx
luizstacio: Not really as the first one only checks for expected states of error I add a new implementation to check if the error is timeout if is not will throw the err.

packages/app/playwright/mocks/database.ts
```diff
@@ -126,11 +129,13 @@ export async function mockData(
             reject(err);
           }
           localStorage.setItem('fuel_isLogged', JSON.stringify(true));
+          localStorage.setItem('password', password);
```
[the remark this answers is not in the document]
luizstacio: Yes this is used only for test even if the user does this on production by hand on the localStorage it will not work.

packages/app/src/systems/Core/services/core.ts
```diff
@@ -0,0 +1,11 @@
+import { db } from '../utils/database';
+import { Storage } from '../utils/storage';
+
+export class CoreService {
+  static async clear() {
+    await db.clear();
+    if (typeof localStorage !== 'undefined') {
+      await Storage.clear();
+    }
+  }
```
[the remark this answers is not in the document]
luizstacio: I was using this on the Background service also this is why implemented this check but is not needed anymore. Going to remove it

packages/app/src/systems/DApp/hooks/useConnectRequest.tsx
```diff
@@ -1,90 +1,80 @@
...
     return state.context.selectedAddresses;
   },
 };
 
 export function useConnectRequest() {
   const { account, accounts, isLoading } = useAccounts();
-  const connectionService = useInterpret(connectMachine);
-  const isConnecting = useSelector(connectionService, selectors.isConnecting);
+  const service = store.useService(Services.connectRequest);
+  const isConnecting = useSelector(service, selectors.isConnecting);
```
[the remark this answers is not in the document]
luizstacio: No need for that as I'm using the store.useService, the result is the same.

packages/app/src/systems/DApp/machines/transactionMachine.test.tsx
```diff
@@ -1,76 +1,57 @@
...
   });
 
   afterEach(() => {
     service.stop();
   });
 
-  it('should approve/send transaction', async () => {
-    await waitFor(service, (state) => state.matches('idle'));
-    service.send('START_REQUEST', {
+  it.only('should approve/send transaction', async () => {
```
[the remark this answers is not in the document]
luizstacio: Don't understand the comment

luizstacio:
> 1- x button is not centralized 2- bold action phrase could be a separated paragraph, or even better could use fuel-ui Alert with warning status.
> 
> image

About the close button, it's an issue on fuel-ui. We can solve it latter https://github.com/FuelLabs/fuel-ui/issues/210.

packages/app/src/systems/DApp/hooks/useConnectRequest.tsx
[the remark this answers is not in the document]
LuizAsFight: Ok, the reason of this comment:
1- we should have a pattern of consuming state from global machines... right now some places use `store.useSelector(Services.connectRequest, selectors.error);` and other use only `useSelector` (from xstate)
2- could there be any difference between this 2 methods? as one of them is implemented by xstate, and the other by our store...

this can led to unexpected data problems that can drain our time debugging until we figure out it's because of different ways of getting data from global state.

you think using on this 2 different ways can bring us problems @pedronauck ??

packages/app/src/systems/DApp/machines/transactionMachine.test.tsx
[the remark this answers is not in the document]
LuizAsFight: I see you removed. all good.

safety means someone might add other tests without realizing this `it.only` resulting on some tests not running

packages/app/src/systems/Settings/machines/setingsMachine.test.tsx
luizstacio: I add the tests to check if the password was updated thanks for the comment

packages/app/src/systems/DApp/hooks/useConnectRequest.tsx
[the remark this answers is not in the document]
luizstacio: Looking at the code of useService you notice it does the same thing. The main difference is that useSelector has both in one action, in places we use a lot of useSelectors I think the best option is todo the useService as it reduces the number of hooks also.

PR #533, opened 2023-02-01
fix: link fuel-ui broken; add support to link fuel-ts

LuizAsFight:
### Linking local dependencies

This PR 
- fix the problem with `fuel-ui` linking
- make it more dynamic to support both `fuel-ui` and `fuel-ts` linking
- Update docs with new instructions

------------------------

For that, `scripts/deps.sh` was created
  - `LINK_FUEL_UI=true` to enable `fuel-ui` linking
  - `LINK_FUEL_TS=true` to enable `fuel-ts` linking
  - it execute the commands to link dependencies
  - runs dev server with local dependencies linked
  - we don't need to use regex. The dependencies are coming from a env var with the exact name, to use regex would need to build a dynamic one which can be trick because of different characters that can be inside input `LINK_DEPS` var


arboleya:
The regex for matching only `fuel-ui/*` packages should look something like this:

```ts
alias: [{
  find: /(@fuel-ui\/[^\s]+)/,
  replacement: path.resolve(__dirname, '../node_modules/$1/dist/index.mjs'
}],
```

Live demo/playground:
 - https://regexr.com/77cf3

LuizAsFight:
> The regex for matching only `fuel-ui/*` packages should look something like this:
> 
> ```ts
> alias: [{
>   find: /(@fuel-ui\/[^\s]+)/,
>   replacement: path.resolve(__dirname, '../node_modules/$1/dist/index.mjs'
> }],
> ```
> 
> Live demo/playground:
> 
> * https://regexr.com/77cf3

THIS PR removes the need to use regex in vite.base.config, as we know exactly the name of deps, just need to iterate over them

arboleya:
IMO, regexes are ideal for string match/replace, providing a smaller code footprint. I am fond of it. :)

Would you mind sharing why you prefer manually iterating over pre-configured environment variables?

LuizAsFight:
> IMO, regexes are ideal for string match/replace, providing a smaller code footprint. I am fond of it. :)
> 
> Would you mind sharing why you prefer manually iterating over pre-configured environment variables?

Yes I agree regex are ideal for that, but for this use case `match/replace` is not needed.

we want to make sure we only apply aliases for dependencies that are being linked locally.

With regex we open room to find other libs, and even if we build a more strict regex, we don't really need to find anything in string, we just need to build a path with the exact dependency name the user provided

for example:
Let's say I have 4 libs:
- @fuel-ui/css
- @fuel-ui/react
- @fuel-ui/test-utils
- @fuel-ui/config

If I only want to link `@fuel-ui/react`, it should be possible

arboleya:
One last example of a strict and flexible regex:
 - `/(@fuel-ui\/(react|css|config|test-utils))/` (_[demo](https://regexr.com/77dqp)_)

It could have many variations:
 - `/(@fuel-ui\/(react))/`
 - `/(@fuel-ui\/(react|css))/`
 - `/(@fuel-ui\/(react|css|config))/`
 - `/(@fuel-ui\/(react|css|config|test-utils))/`
 - `/(@fuel-ui\/(react|css|config|test-utils|new-lib-from-the-future))/`

Even these seem more concise than a mix of env vars and manual iteration.

Feel free to use the approach you think is best.

Over and out. 🫡

LuizAsFight:
@arboleya I added support to `fuel-ts` as well now

the dependencies come into node front-end process in `LINK_DEPS` env var, because we reuse them from prev command executed in `deps.sh`

if we want a regex to process it, would need to build it dynamically, which can be tricky because of characters inside dependency name that would need to be escaped (`@` `/` ` `white space, others?)

Issue #308, opened 2022-12-02
Improve the code injection of FuelWeb3 on window


luizstacio:
## Description

When a developer tries to use `window.fuel`, it can get undefined. The cause of the problem is the time between the page being loaded and the `contentScript` is executed. The idea here is to trigger an event on the window instance, after the script is loaded, enabling the application to correctly detect when is loaded.

## Ideias on how to implement

- On CRX `contentScript.ts` add inside the `script.onload` event a dispatch of the event `fuelLoaded`.
- Change the docs `useFuel` examples to use the new `fuelLoaded` event.
- Add a test on the crx.test.ts blank page to see if the event is triggered.
- Add to the docs how to listen when `window.fuel` is loaded.

PR #541, opened 2023-02-04
feat: improve fuel injection

luizstacio:
This PR improves how we detect when fuel is injected on the page. With this PR now, it's possible to listen for an event on the document object `FuelLoaded` this name makes reference to `DOMContentLoaded`. This event is triggered right after the fuel object is assigned to the target object, in this case, the window.

This PR also adds a new handler to `onInstall` event. The new handler injects `fuel` into the tabs already open, removing the need to refresh the page.

These two improvements enable applications to listen to the event `FuelLoaded`, and once the user installs the Wallet, the application can start work without the need for workarounds or request for the user to refresh the page.

close: #308


packages/docs/docs/how-to-use.mdx
```diff
@@ -14,6 +14,15 @@ window.fuel.connect();
 
 You can try this code directly in the developer console.
 
+## Detect when fuel is loaded
+
+It's possible that your applaction loads before `window.fuel` is injected. To detect when the `fuel` is loaded and ready to use, you can listen to the event `FuelLoaded` on the document.
```
tomiiide, line 19:
```suggestion
It's possible that your application loads before `window.fuel` is injected. To detect when the `fuel` is loaded and ready to use, you can listen to the event `FuelLoaded` on the document.
```

tomiiide (changes requested):
Looks good, just one little typo.

tomiiide (approved):
Looks good now.

packages/docs/examples/FuelLoaded.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+    // If fuel is already loaded, call the handler
+    if (window.fuel) {
+      onFuelLoaded();
+    }
+
+    // Listen for the fuelLoaded event
+    document.addEventListener('FuelLoaded', onFuelLoaded);
+
+    // On unmount, remove the event listener and clear the timeout
```
eswarasai, line 23:
```suggestion
    // On unmount, remove the event listener
```

LuizAsFight (approved):

LuizAsFight (changes requested):
tests breaking after merging with master

Issue #531, opened 2023-02-01
Improve store performance and implementation


pedronauck:
### Description

The implementation of our store is working very well, but there's a bunch of space for improvements that could bring us better performance and even avoid some implementation issues.

#### Some of the possible improvements

- [ ] Be able to persist state ([something like this](https://github.com/FuelLabs/fuels-wallet/issues/530))
- [ ] Check implementation to avoid imports order being an issue
- [ ] Improve performance to avoid re-renders
- [ ] ...

PR #544, opened 2023-02-06
fix: refactor on store to fix bugs related to it + improvements

pedronauck:
*tdlr;*

⭐ Fix #376 #531 

# Introduction
This pull request aims to improve the current global xstate machines system (aka xstore) in our application. The current system is facing several problems such as cycle dependencies, unexpected bugs, and difficulty in tracking and debugging issues. To overcome these challenges, a new approach has been proposed using Jotai as a global store to manage the machines outside the React scope. The new approach not only brings performance improvements but also reduces the chances of encountering unexpected bugs.

## Main problems
- Cycle dependencies are causing difficulties in tracking issues in the application.
- Unexpected bugs related to runtime variables are arising due to the recreation of classes on live-reloading.
- Services are being started multiple times leading to performance issues.
- The current non-reactive implementation is making it hard to catch bugs.

## Solution
- To solve these problems, a new `StoreProvider` has been introduced in the application using [Jotai](https://jotai.org/) behind the scenes as a global store to provide a safer and more robust solution, giving more control and predictability. Additionally, the new approach has introduced several APIs to the store and improved the overall performance of the application and now localStorage is used for cache state with an option to blacklist a state.

##### `store.addMachine(key, () => machine)`
This API allows adding a new state machine to the store.

##### `store.addHandlers((store: Store) => Handlers)`
This API allows adding handlers for state management.

##### `store.onStateChange(key | Service, (state: State) => void)`
This API allows listening to changes in state.

##### `store.onStartStore(() => void)`
This API allows executing logic when the store starts.

##### `store.getStateFrom(key)`
This API allows get some state from a specific service

##### `store.waitFor(key, string | (state) => boolean)`
This API allows waiting for some service state using the store

##### `const [state, send] = store.useState('accounts')`
This API allows using the state of the store in components.

## LocalStorage cache

As mentioned now it's possible to have a cache in some states using LocalStorage to get the initial state when service starts.  This also possible to control which states you want to be saved or not by using `storageBlacklist` prop:

### Usage

```tsx
import { createStore } from '@fuel-wallet/xstore';

export enum Services {
  accounts = 'accounts',
  networks = 'networks',
}

export type StoreMachines = {
  accounts: AccountMachine;
  networks: NetworksMachine;
};

export const store$ = createStore<StoreMachines>({
  id: 'fuelStore',
  persistedStates: ['networks'],
});

export const store = store$
  .addMachine(Services.accounts, () => accountMachine)
  .addMachine(Services.networks, () => networksMachine)
  .addHandlers(accountEvents)
  .addHandlers(networkEvents)
  .setup();

export const { StoreProvider } = store;
```

### Pros
- The integration of these APIs enhances the application's performance and simplifies state management, reducing the risk of errors. This leads to increased predictability and fewer unexpected bugs.
- Be able to send a dispatch from the store anywhere and outside React context
- Be able to make a broadcast and send an event for all machines
- Simple API to handle things
- Be able to `useSetConfigMachine`
- Don't need to create a dependency between machines (decoupling)

### Cons
- We have a trade-off right now that's a new library in the lower layer of the application. In fact, we could choose to try to implement all in native Javascript *(I tried hard, believe me)* but after days of working on this, was hard to find a better/safer solution that covered all things we need.
- Abstractions that can lead to bugs sometimes

Since [Jotai has just 2.3kb](https://bundlephobia.com/package/jotai@2.0.0) gzip, I don't see this as a big problem for us. I think is worth the trade-off.

> 📝 before someone asks, we have technical requirements that prevent us from using React.Context. Instead, we are able to dispatch actions from the store anywhere in the application by simply importing it and calling `store.send`. This approach would not be possible with React Context, as we would need to constantly inject actions into the state machines.


package.json
```diff
@@ -46,7 +46,7 @@
     "node:dev": "fuelhat node",
     "nodes:start": "fuelhat node start -d && fuelhat node start -t -d",
     "nodes:stop": "fuelhat node stop -d && fuelhat node stop -t -d",
-    "nodes:reset": "pnpm nodes:start && pnpm nodes:stop",
+    "nodes:reset": "pnpm nodes:stop && pnpm nodes:start",
```
tomiiide, line 49: 😂 Nice catch.

luizstacio (changes requested):
I am testing the new implementation

packages/store/src/atoms/createStoreAtoms.ts
```diff
@@ -0,0 +1,243 @@
...
+  const machinesObjAtom = atom<Obj>({} as Obj);
+  const machinesAtom = atom(
+    (get) => get(machinesObjAtom),
+    (get, set, input: AddMachineInput<T, keyof T>) => {
+      const curr = get(machinesObjAtom);
+      const machineAtom = atomWithMachine(input);
+      const machinesObj = { ...curr, [input.key]: machineAtom };
+      set(machinesObjAtom, machinesObj);
+      return machinesObj as Obj;
+    }
```
luizstacio, line 68: Improve the code here by splitting it set / get into a function and passing here by name, it will help to understand what each part does.

luizstacio: This could be done for every new atom been created on this file.

luizstacio: And for every callback function

packages/store/src/atoms/atomWithSubscription.ts
```diff
@@ -0,0 +1,53 @@
...
+    }
+  );
+  subscriptionAtom.onMount = (init) => {
+    init({ type: 'init' });
+    return () => {
+      subs.forEach((sub) => sub.unsubscribe());
+      subs.clear();
+    };
+  };
+  return subscriptionAtom;
```
luizstacio, line 52: Improve code spliting and callback function names.

packages/store/src/utils/storage.ts
```diff
@@ -0,0 +1,31 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
luizstacio, line 1: We already implement this here: `packages/app/src/systems/Core/utils/storage.ts`. As I notice the code is replicated can we create a package utils, can be done on another issue.

luizstacio (changes requested):
I have tested the new implementation, and it looks smooth. Not found issues.

I left some reviews to improve code readability.

Nice work!

LuizAsFight (changes requested):

packages/store/src/atoms/atomWithSubscription.ts
pedronauck: I improved this type of atom to be more concise, I believe it's easier to understand right now.

packages/store/src/atoms/createStoreAtoms.ts
pedronauck: I create a file for each atom factory we have inside the store, about the callbacks, which will be kinda unnecessary and we'll lose a lot of type inference by doing this. It's not hard to understand, it's the way Jotai works. An atom can have this signature:

```tsx
// primitive atom
function atom<Value>(initialValue: Value): PrimitiveAtom<Value>

// read-only atom
function atom<Value>(read: (get: Getter) => Value | Promise<Value>): Atom<Value>

// writable derived atom
function atom<Value, Update>(
  read: (get: Getter) => Value | Promise<Value>,
  write: (get: Getter, set: Setter, update: Update) => void | Promise<void>
): WritableAtom<Value, Update>

// write-only derived atom
function atom<Value, Update>(
  read: Value,
  write: (get: Getter, set: Setter, update: Update) => void | Promise<void>
): WritableAtom<Value, Update>
```

Basically, when an atom has just the first callback, it's just a read-only atom, when both it's a writable atom (can read, and write).

packages/store/package.json
```diff
@@ -23,15 +23,26 @@
...
+    "fast-equals": "^4.0.3",
+    "jotai": "^2.0.0",
+    "pnpm": "^7.26.3",
     "react": "^18.2.0",
+    "uuid": "^9.0.0",
     "xstate": "^4.35.4"
   },
   "devDependencies": {
+    "@fuel-wallet/config": "workspace:*",
+    "@fuel-ui/test-utils": "^0.12.0",
```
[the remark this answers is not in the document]
pedronauck: ✅

packages/store/src/utils/storage.ts
pedronauck: I added it as a new package, but didn't change inside the `app` to avoid bloat this pull request more and more 😅 but it's implement in the store, so I can change inside the app in another PR

luizstacio (approved):

LuizAsFight (approved):

Issue #478, opened 2023-01-24
Change home network selector to use dropdown layout


luizstacio:
### Description

When clicking on the home network tab, the user should be able to change between networks quickly. Currently, we are redirecting the user to the network page.

We should replace this behavior to show the dropdown selector equal to the behavior on the sidebar network selector.

### Implementation ideas
- Move NetworkSelector to the network system components
- Change the network component usage on the sidebar to use the new path
- Change the behavior of the Home page network button to open the NetworkSelector

PR #546, opened 2023-02-07
feat(app): updated top nav to use network dropdown

eswarasai:
### Description
- [x] Change home network selector to use dropdown layout
- [x] Removed Network switcher from Sidebar
- [x] Added indicator for selected network

Fixes: #478 

### Screenshot
network-dropdown

### TODO
- [ ] Support `Add Network` option in the Dropdown Menu (refer to below screenshot)
add-network


luizstacio:
> 

I think add network can be split in a new issue, I don't see the need for this right now.

luizstacio (changes requested):
When use has only one network, the dropdown should open, this makes the behavior consistent, and the user can understand how the feature works.

tomiiide:
One small enhancement.

The dropdown auto focuses the first network in the list. 
I think the selected network should have the default focus, or an indicator that it is selected.

Screenshot 2023-02-07 at 3 15 28 PM
Screenshot 2023-02-07 at 3 14 56 PM

tomiiide (changes requested):
Great work so far.
 
 Just a small enhancement requested.

luizstacio (changes requested):
Screenshot 2023-02-07 at 18 37 33

The indicator is making the layout unaligned. This was a good idea. To fix this, we can use the same style we implement on account and network lists. The little green border.

Screenshot 2023-02-07 at 18 39 01

LuizAsFight (changes requested):
This is the wallet's header when I first open the network dropdown, with more than 1 network.
image

it's looking good

-------------
we can improve some things though
- dropdown has no highlight over the rest of content, looks mixed 
  - for that we can insert a shadow, for example `box-shadow: 0 2px 3px 4px rgb(0 0 0 / 20%)`.. giving a sense of the dropdown being over the rest of content
image
- an even better approach would make this dropdown in a modal style, with a slight black overlay over the rest of content
image

- first network comes in a "hover" style, and never goes away until I click outside
  - is there a way to apply that styly only when hovering that element?
  

https://user-images.githubusercontent.com/8636507/217646786-62322515-b01c-4e53-85e9-7cfd042af5ad.mov

luizstacio (approved):

eswarasai:
@LuizAsFight - Added `box-shadow` to Dropdown Menu like suggested above and also fixed the issue with `autoFocus` not focusing on selected network item. Need to merge this [PR](https://github.com/FuelLabs/fuel-ui/pull/206) and publish a new version of `fuel-ui` package before we can merge this PR. Thanks!

cc: @luizstacio

packages/app/src/systems/Core/components/Layout/Layout.tsx
```diff
@@ -124,6 +124,7 @@ export const styles = {
     height: WALLET_HEIGHT,
     background:
       'linear-gradient(210.43deg, #0E221B 0%, #071614 10.03%, #0C0E0D 18.38%)',
+    zIndex: '$0',
```
luizstacio, line 127: Why we need this here?

luizstacio (approved):

LuizAsFight (approved):

packages/app/src/systems/Core/components/Layout/Layout.tsx
eswarasai: Since the dropdown has the `zIndex: '$1'`, to avoid it being present on top of the overlay content (refer to below screenshot), setting the layout to be at `zIndex: '$0'`

network-dropdown

tomiiide (approved):
looks good to me.

PR #526, opened 2023-01-31
docs: add events examples to the documentation

tomiiide:
### what does this PR do?
This PR add creates an Enum that is a list of all events the fuel clients emits.

It also add examples of how to listen to the events to the docs.

## how to test?
Visit `docs/api/#events/`
https://fuels-wallet-ikg8evbyr-fuel-labs.vercel.app/docs/api/#events

closes #422

PR #551, opened 2023-02-08
refactor: change `fuel.on` to accept string and externalize events on fuel instance

luizstacio:
The implementation made on PR #526 has added events as an enum, but with this change, the user had to always import the Enum from the package in order to use it. The idea here is to externalize this enum directly on the fuel instance on the window.

### DIFF
--- a/.changeset/tough-nails-compete.md
+++ b/.changeset/tough-nails-compete.md
@@ -1,0 +1,6 @@
+---
+'@fuel-wallet/types': patch
+'@fuel-wallet/sdk': patch
+---
+
+Move events enum to fuel object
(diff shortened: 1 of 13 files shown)


tomiiide (dismissed):
LGTM, nice improvement.

tomiiide (approved):

tomiiide (changes requested):
build is breaking

tomiiide:
Build was breaking  some `CodeImport` elements in the markdown examples was invalid. The code in the start and end lines wasn't a valid code block.

I updated the start and end lines, and build is successful.

tomiiide (approved):

Issue #504, opened 2023-01-24
Inactivity error when sending transaction


FossilFrank:
### What version of Wallet are you using?

0.0.1

### What is happening?

When trying to send a transaction, the window will pop up, and immediately close. I receive the error "Window closed by inactivity after 5 minutes!"

### steps to reproduce

Can be reproduced using this CodeSandbox:

https://codesandbox.io/s/microchain-deploy-z4lkcr

### How it should behave?

Should prompt to sign the transaction

### What os are you seeing the problem on?

Mac/Apple Desktop

### What browser?

Chrome

luizstacio:
@FossilFrank can you include de ABIs on the code sandbox, the current example is broken because of that.

FossilFrank:
@luizstacio updated!

chappie1998:
I am also getting the same error.

luizstacio:
@FossilFrank and @chappie1998, thanks for the reports. I found the issue inside `fuels-ts` we are working to fix it, and soon this PR https://github.com/FuelLabs/fuels-ts/pull/766 will be merged. We are going to update this on the Fuel Wallet.

luizstacio:
@FossilFrank  and @chappie1998, we have released a new version of fuels@0.30.0 this fixes part of the issue but when the tx fails the error is not returning also this is going to be fixed once #555 is merged.

Also updated the project you sent with the last versions: https://codesandbox.io/s/microchain-deploy-forked-izs12h

chappie1998:
@luizstacio after updating with fuels 0.31.0, I am still facing same issue.

PR #555, opened 2023-02-09
fix: improve return error on failed txs

luizstacio:
- Update fuels-ts to support new tx stringify
- Improve return error on tx simulate

close: #504


eswarasai (approved):
LGTM 👍

PR #558, opened 2023-02-10
feat: Custom Assets Management

### LINKED ISSUE #479: Create Asset management tool
The nature of the UTXO enables the wallet to list all assets from a given account, but now the Wallet only lists the known assets from a hardcoded list, making it impossible to see other assets.

To improve this, we should create a way for adding custom assets metadata;

Where;
- Users add custom tokens to the wallet, making their UTXOs with unknown assets ids been shown on the Wallet, the metadata required is;
  - Name
  - Symbol
  - Image
  - Asset Id
- DApp developer to request adding assets to the user's Wallet.

Note: Also consider on this feature a fetch to a static json containing tokens metadata.

LuizAsFight:
# Add Asset Management flow

### Added Assets List screen, it has 2 tabs
`Listed Assets`
- the initial assets the wallet already have
- User can't edit from the list

`Custom Assets`
- User can add / edit from the list
- Assets that the user added/edited

`Assets = Listed Assets + Custom Assets`

### Wallet homepage (Balances)
If user has some assetId that is not in `Assets` list, we will show as an `Unknown Asset` , and add a CTA to create a new `Custom Asset` from there

### Created screen to DAPPs add `Custom Assets` in the wallet using the sdk
Now developers can use `fuel.addAsset` to request the user to insert assets

### Create `fuel.on('assets')` to listen to assets changes

Closes #479


luizstacio (changes requested):
- On adding an asset inside the wallet after adding an asset it should keep the tab custom assets open it's redirecting to the assets list.
- The empty screen on custom assets didn't have a nice layout is missing some borders I think we should follow the same design of other screens with cards and the list would take all the space it will solve this issue and keep the design consistent
- SDK addAsset, we could add an asset line preview and add some text describing what is going to happen.
- I tried to do a mint request but the TX screen stayed on loading forever.

Screenshot 2023-02-13 at 16 08 38

Screenshot 2023-02-13 at 15 52 45
Screenshot 2023-02-13 at 16 19 17
Screenshot 2023-02-13 at 16 17 20

packages/assets-listed/src/index.ts
```diff
@@ -0,0 +1,22 @@
...
+    symbol: 'DAI',
+    imageUrl: `assets/dai.svg`,
+  },
+  {
+    assetId: '0x2260fac5e5542a773aa44fbcfedf7c193bc2c599',
+    name: 'Bitcoin',
+    symbol: 'BTC',
+    imageUrl: `assets/btc.svg`,
+  },
+];
```
luizstacio, line 22: We can remove DAI and Bitcoin for now as we don't support this assets, keep just ETH for now.

packages/assets-listed/package.json
```diff
@@ -0,0 +1,6 @@
+{
+  "name": "assets-listed",
```
luizstacio, line 2: The name of the package could be `asset-list`

packages/docs/examples/AddAsset.tsx
```diff
@@ -0,0 +1,99 @@
...
+  const [fuel, notDetected] = useFuel();
+  const [isConnected] = useIsConnected();
+  const [addedAsset, setAddedAsset] = useState<Asset>();
+  const [asset, setAsset] = useState<Asset>({
+    name: 'New',
+    symbol: 'NEW',
+    assetId:
+      '0x566012155ae253353c7df01f36c8f6249c94131a69a3484bdb0234e3822b5d90',
+    imageUrl:
+      'https://upload.wikimedia.org/wikipedia/commons/thumb/4/46/Bitcoin.svg/1200px-Bitcoin.svg.png',
```
luizstacio, line 22: Can we add this asset config to some mock file and reuse it in all places?

packages/docs/examples/AddAsset.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+  return (
+    <ExampleBox error={errorMessage}>
+      <Stack css={{ gap: '$4' }}>
+        <Input isDisabled={!isConnected} css={{ width: '100%' }}>
+          <Input.Field
+            value={asset.assetId}
+            onChange={(e) => setAsset({ ...asset, assetId: e.target.value })}
+            placeholder="Type your assetId (0x...)"
+            // css={{ color: '$whiteA11', padding: '$2' }}
```
luizstacio, line 46: Remove comment props

packages/docs/examples/Transfer.tsx
```diff
@@ -29,16 +29,19 @@ export function Transfer() {
   const [addr, setAddr] = useState<string>(
     'fuel1a6msn9zmjpvv84g08y3t6x6flykw622s48k2lqg257pf9924pnfq50tdmw'
   );
+  const [assetId, setAssetId] = useState<string>(
+    '0x0000000000000000000000000000000000000000000000000000000000000000'
```
luizstacio, line 33: Can use `NativeAssetId` from `fuels` here.

packages/docs/examples/events/index.tsx
```diff
@@ -1,4 +1,5 @@
 import { Accounts } from './Accounts';
+import { AssetsExample as Assets } from './Assets';
```
luizstacio, line 2: Maybe just rename the class?

packages/app/src/systems/Account/hooks/useAccounts.tsx
```diff
@@ -49,12 +50,21 @@ const listenerAccountFetcher = () => {
...
   const account = store.useSelector(Services.accounts, selectors.account);
   const overlay = useOverlay();
+  const balanceAssets = useMemo(
+    () =>
+      account?.balances?.map((balance) => ({
+        ...balance,
+        ...(assets?.find(({ assetId }) => assetId === balance.assetId) || {}),
+      })),
+    [account?.balance, assets]
+  );
```
luizstacio, line 67: Should we move this code to inside the machine?

packages/assets-listed/src/index.ts
```diff
@@ -0,0 +1,22 @@
+export const ASSETS_LISTED = [
```
luizstacio, line 1: I think we can export as `AssetList`.

packages/app/src/systems/Asset/components/AssetItem/AssetItem.tsx
```diff
@@ -1,48 +1,141 @@
...
-  );
+  if (isHidden) return null;
+  const {
+    symbol,
+    name = '',
+    imageUrl,
+    amount,
+    assetId,
+    isCustom,
+  } = asset || {};
```
luizstacio, line 56: We should move the isHidden to an upper component and filter the hidden assets to avoid unnecessary render.

packages/app/src/systems/CRX/background/services/BackgroundService.ts
```diff
@@ -271,4 +274,33 @@ export class BackgroundService {
...
+
+  async addAsset(
+    input: Exclude<MessageInputs['addAsset'], 'origin'>,
+    serverParams: EventOrigin
+  ) {
+    const origin = serverParams.origin;
+
+    const app = await ConnectionService.getConnection(origin);
+    if (!app) {
+      throw new Error('Unable to establish a connection. No accounts found');
```
luizstacio, line 292: This is already validated by the middleware you don't need to check connection again here.

packages/app/src/systems/DApp/machines/addAssetMachine.tsx
```diff
@@ -0,0 +1,141 @@
...
+           * to avoid stuck black screen, should close the window and let user retry */
+          TIMEOUT: '#(machine).closing', // retry
+        },
+      },
+      closing: {
+        entry: ['closeWindow'],
+        always: {
+          target: '#(machine).failed',
+        },
+      },
```
luizstacio, line 61: This is not required anymore, we have implemented a closePopUp when the connection is lost (user refresh)

luizstacio (changes requested):

packages/docs/examples/AddAsset.tsx
LuizAsFight: added mock in `playwright/e2e` and `src/app/assets/mock`

packages/app/src/systems/Account/hooks/useAccounts.tsx
LuizAsFight: no because we would need to inject `assets` in machine context, it's not needed as it comes from `useAssets`.

I will put this code in a closure selector instead

packages/app/src/systems/Asset/components/AssetItem/AssetItem.tsx
LuizAsFight: in `AccountItem` it's done in the exact same way, if you want we can open an issue to fix it there later...

I will fix only in `AssetItem` for now

luizstacio (approved):

PR #564, opened 2023-02-15
refactor: increase max listeners and create config file on SDK

luizstacio:
This PR increases the max number of listeners before showing a warning of a memory leak, and also moves the consts we have to a config file.


tomiiide (approved):
looks good.

Issue #563, opened 2023-02-15
Move Networks from Page to Modal


eswarasai:
### Description
Transform the flow for listing, adding and editing Networks to open in a modal instead of a page, to keep consistent with the Accounts UX

PR #570, opened 2023-02-16
feat(app): add networks as modal instead a route

eswarasai:
Fixes: #563


packages/app/src/systems/Network/hooks/useNetworks.ts
```diff
@@ -39,33 +34,28 @@ export function useNetworks(opts: NetworkInitialInput = {}) {
...
-  }, [opts.networkId, opts.type]);
-
   return {
     handlers: {
+      closeDialog,
       goToUpdate,
-      goToAdd,
-      goToList,
+      goToAdd: store.openNetworksAdd,
+      goToList: store.openNetworksList,
```
luizstacio, line 58:
```suggestion
      openNetworksAdd: store.openNetworksAdd,
      openNetworks: store.openNetworksList,
```

luizstacio (changes requested):
Just small changes

luizstacio (dismissed):

packages/app/src/systems/Network/hooks/useNetworks.ts
```diff
@@ -39,37 +34,32 @@ export function useNetworks(opts: NetworkInitialInput = {}) {
...
-  }
-  function goToAdd() {
-    navigate(Pages.networkAdd());
+  function closeDialog() {
+    overlay.close();
   }
-  function goToList() {
-    navigate(Pages.networks());
+  function goToUpdate(network: Network) {
+    store.editNetwork({ id: network.id as string });
```
LuizAsFight, line 49:
```suggestion
  function goToUpdate(id: string) {
    store.editNetwork({ id });
```

packages/app/src/systems/Network/pages/Networks/Networks.tsx
```diff
@@ -1,38 +1,58 @@
...
+
+const styles = {
+  description: cssObj({
+    ...coreStyles.scrollable('$gray3'),
+    padding: '$4',
+    flex: 1,
+
+    '&[data-has-scroll="true"]': {
+      padding: '$4 $2 $4 $4',
+    },
```
LuizAsFight, line 56: imo this padding should be applied by default, not only in "scroll mode"

packages/app/src/systems/Network/pages/Networks/Networks.tsx
```diff
@@ -1,38 +1,58 @@
...
+          variant="link"
+          icon={<Icon icon="X" color="gray8" />}
+          aria-label="Close networks modal"
+          onPress={handlers.closeDialog}
+        />
+      </Dialog.Heading>
+      <Dialog.Description
+        as="div"
+        css={styles.description}
+        data-has-scroll={Boolean((networks || []).length >= 6)}
```
LuizAsFight, line 25: setting `data-has-scroll` to increase padding could be removed, and apply the style always instead.

defining scroll boolean based on networks length is fragile because any changes in networkItem element will change height and the dev would need to come back here to change length to scroll...

LuizAsFight (changes requested):
looking good, minor comments

LuizAsFight:
in Figma we have these nice circle status showing if network is online

@luizstacio should we introduce them in this PR?
if not, do we have an issue to do it later ?

we only have green in design by now, but we can also include yellow for `slow` networks (let's say more than 1 second delay)

image

luizstacio:
> in Figma we have these nice circle status showing if network is online
> 
> @luizstacio should we introduce them in this PR? if not, do we have an issue to do it later ?
> 
> we only have green in design by now, but we can also include yellow for `slow` networks (let's say more than 1 second delay)
> 
> image

We remove the dot because it don't represent nothing right now.

packages/app/src/systems/Network/components/NetworkItem/NetworkItem.tsx
```diff
@@ -37,7 +37,7 @@ export const NetworkItem = forwardRef<HTMLDivElement, NetworkItemProps>(
             variant="link"
             icon={<Icon icon={Icon.is('Pencil')} />}
             aria-label="Update"
-            onPress={() => onUpdate?.(network)}
+            onPress={() => onUpdate?.(network.id as string)}
```
LuizAsFight, line 40: enforcing a type to match function input typing is basically throw away typescript benefits and open a door to break the code

instead we can change `onUpdate` function to accept undefined

packages/app/src/systems/Network/components/NetworksDialog/NetworkDialog.stories.tsx
```diff
@@ -0,0 +1,16 @@
+import { Box } from '@fuel-ui/react';
```
LuizAsFight, line 1: checking PR storybook, nothing is shown in this story.

can we add a button to open dialog?

LuizAsFight (changes requested):

packages/app/src/systems/Network/components/NetworksDialog/NetworkDialog.stories.tsx
eswarasai: Can you check if the storybook for `NetworksDialog` is rendering fine for you? With `viewport` parameters configured for a component, I'm getting errors locally when trying to view it's storybook.

LuizAsFight (changes requested):
storybook for NetworkDialog still broken

before was empty...
now it throws error on render, white screen  as result and this error in console 
image

luizstacio (approved):

PR #581, opened 2023-02-21
feat: change checkboxes on signup mnemonic phrase screen

### LINKED ISSUE #465: Improve checkbox layout on backup phrase confirmation
Currently, the checkbox is not bringing enough attention, and the content is very poor. Because of that, we should change the layout and text of the confirm copy checkbox.

Add two checkboxes text:

- I have copied my recovery phrase in a safe place, and I have access to it to restore my wallet.

For the checkbox, we should use the warning box with a tag in yellow, for example:
Screenshot 2023-01-22 at 17 43 48

LuizAsFight:
Closes #465

### DIFF
--- a/packages/app/playwright/crx/crx.test.ts
+++ b/packages/app/playwright/crx/crx.test.ts
@@ -116,7 +116,10 @@ test.describe('FuelWallet Extension', () => {
 
       /** Copy Mnemonic */
       await getButtonByText(page, /Copy/i).click();
-      await page.getByRole('checkbox').click();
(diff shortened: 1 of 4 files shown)


eswarasai:
Thoughts on adding a little margin-top for the checkboxes content, something like below?
Maybe adding a little margin between both the checkboxes might highlight them better?

checkboxes

LuizAsFight:
@eswarasai looking to your screenshot it makes sense. just did the change can you check again ?

eswarasai (approved):
LGTM 👍

PR #593, opened 2023-02-24
feat: support insert multiple assets via SDK fuel.addAsset

### LINKED ISSUE #584: SDK addAssets enable DApps to request add multiple assets
Currently the SDK provides a method `addAsset` this receives a single asset per time, which for some applications can be tedious as they need to add multiple assets to the user wallet.

In this work, we need to;
- Add a method `addAssets` that receives a array of asset metadata
- On Approve Add Asset screen show the list of assets been added
- If some of the assets conflict with assets already on the Wallet, ignore the asset.

### LINKED ISSUE #598: Add Asset page is not blocking same assetId

0.0.1

### What is happening?

When adding an asset manually inside the Wallet. Assets page, add a new asset, add an asset_id that already exists, add name and symbol, and click on save. The Wallet just updates a current asset without alerting it that already exists and blocking the action

### steps to reproduce

1- click Menu -> settings -> Assets
2- click Add Asset
3- add an asset with same assetId that was already created before
4- click save

### How it should behave?

The behavior would be to blocking the action showing a error

None

### What browser?

None

_No response_

LuizAsFight:
Closes #584 
Closes #598


luizstacio (changes requested):
I found two things;
1. When adding an asset manually inside the Wallet. Assets page, add a new asset, add an asset_id that already exists, add name and symbol, and click on save. The Wallet just updates a current asset without alerting it that already exists and blocking the action.
2. On the origin info is showing the wrong domain.

Screenshot 2023-02-26 at 22 59 56

LuizAsFight:
> I found two things;
> 
> 1. When adding an asset manually inside the Wallet. Assets page, add a new asset, add an asset_id that already exists, add name and symbol, and click on save. The Wallet just updates a current asset without alerting it that already exists and blocking the action.
> 2. On the origin info is showing the wrong domain.
> 

About point 1. I think we can do in another issue / PR as it's orthogonal to this issue/PR objective
https://github.com/FuelLabs/fuels-wallet/issues/598

luizstacio (changes requested):

luizstacio (approved):

Issue #486, opened 2023-01-24
Move TX Request from Page to Modal


luizstacio:
When approving a transaction, the user is redirected to a new page, making it lose the application state. We should transform this flows to open the transaction approval flow on a modal instead of a page.

PR #597, opened 2023-02-27
feat(app): add approve transaction as modal

eswarasai:
Fixes: #486 

### TODO:
- [x] Add storybook for `TxApprove` component
- [x] Refactor Accounts, Network and Transaction dialog into single component (unnecessary duplication of styles etc.)


luizstacio (changes requested):
Code looks good!

Some issues found;
- When doing a TX
  - The popup opens without showing the tx loader and than to content popups up, we should show the tx loader
   - Inside the popup, the content has a different color than the background of the regular popups, making the cards mix with the background.
    - Right after clicking on the send if I click on cancel inside the approval tx, the popup shows a tx loader, and I need to click again on cancel to close the popup.
    - After confirming the TX not shows any state of loading and looks like it had failed if I click again then it works.
    
    For testing these behaviors I have used throttling on the network set as `slow 3g`.
    
    https://user-images.githubusercontent.com/3941923/222194380-996ddb04-df0f-415e-9d51-dbb21988481b.mp4

packages/app/src/systems/Overlay/components/OverlayDialog/OverlayDialog.tsx
```diff
@@ -26,6 +37,7 @@ const styles = {
     height: WALLET_HEIGHT,
     maxWidth: WALLET_WIDTH,
     maxHeight: 'none',
+    background: '#0C0E0D',
```
eswarasai, line 40: > Inside the popup, the content has a different color than the background of the regular popups, making the cards mix with the background.
@luizstacio - To fix the above issue, made the Overlay background darker than `$gray1`, let me know if there's a better color to use here instead.

luizstacio (approved):

Issue #469, opened 2023-01-23
Wallet Setup - Wallet Created Page


luizstacio:
- [ ]  Remove account details
- [ ]  Add warning on the top corner:
    - [ ]  Pin Fuel Wallet extension
    - [ ]  Description: Click on (the icon of the extensions) and (the icon of the pin).
      https://www.figma.com/file/uEckQY4EaQ5rXwyISg6d1g/Fuel-Wallet?node-id=209%3A10186&t=mRnTAw1XuiqfxJ4i-4
- [ ]  Add button
    - [ ]  Finish (closes the signup flow)

PR #602, opened 2023-02-28
chore: update welcome screen texts

luizstacio:
For now, we are keeping account details but adding the warning and a text with instructions to pin the wallet. As we want the user to Pin the Wallet, we didn't add a button to close the tab.

close: #469


packages/app/src/systems/SignUp/components/WalletCreated/components/PinWalletCard.tsx
```diff
@@ -0,0 +1,56 @@
+import { cssObj } from '@fuel-ui/css';
```
LuizAsFight, line 1: about `PinWalletCard` 

- can we move it to `systems/Core` ?
- can we create a simple story for it?

LuizAsFight (changes requested):

packages/app/src/systems/SignUp/components/WalletCreated/components/PinWalletCard.tsx
luizstacio: It doesn't need to be in the core, as it's not supposed to be used elsewhere. About the storybook will create it.

LuizAsFight (approved):

PR #604, opened 2023-02-28
fix: password checker detect more symbols

### LINKED ISSUE #585: Password input not accepts especial chars
When inputting some password with `*-_`, the current input does not accept valid chars, but it should. The input should accept any letter, char, or symbol.

LuizAsFight:
Closes #585


matt-user (approved):
why was the changeset pr workflow skipped?

PR #609, opened 2023-03-03
feat: import account from private key

### LINKED ISSUE #591: feature request: import private key
### Motivation

It would be nice to be able to import a single account using a private key.

### Usage example

This is especially helpful for testing with a local instance of Fuel Node running.

### Possible implementations

Similar to how metamask allows the same functionality.

LuizAsFight:
Closes #591


packages/app/src/systems/Account/hooks/useAccountFormNameObj.tsx
```diff
@@ -0,0 +1,15 @@
+import type { Account } from '@fuel-wallet/types';
+import { useMemo } from 'react';
+import * as yup from 'yup';
+
+export function useAccountFormNameObj(accounts: Account[]) {
```
luizstacio, line 5: Why NameObj on the end?

packages/app/src/systems/Account/machines/accountsMachine.tsx
```diff
@@ -333,14 +363,53 @@ export const accountsMachine = createMachine(
...
+          }
+          if (!input?.name.trim()) {
+            throw new Error('Name cannot be empty');
+          }
+
+          // Add account to vault
+          const accountVault = await VaultService.createVault({
+            type: 'privateKey',
+            secret: input.privateKey,
+          });
```
luizstacio, line 392: You don't need to create a new vault for each privateKey a vault of type privateKey can handle multiple different privateKeys.

packages/app/src/systems/Core/utils/string.tsx
```diff
@@ -21,3 +21,41 @@ export const truncate = (str: string, length: number = 30) => {
...
+    const repeatedAssets = allValues.filter(
+      (a) => a && a.trim() === value.trim()
+    );
+
+    let notRepeatedField: string;
+    if (repeatedAssets[0]) {
+      const nextToTry =
+        tries === 1
+          ? `${value} (${tries})`
+          : `${value?.slice(0, -4)} (${tries})`;
```
luizstacio, line 49: Why we need this slice here?

luizstacio (changes requested):

packages/app/src/systems/Core/utils/string.tsx
LuizAsFight: remove last (1) and inform new (2) for example.

luizstacio: And if the number is `Text (10)` What happens?

LuizAsFight: good catch my friend

packages/app/src/systems/Account/machines/accountsMachine.tsx
luizstacio: https://github.com/FuelLabs/fuels-wallet/issues/617

luizstacio (dismissed):

luizstacio (approved):

PR #619, opened 2023-03-11
feat: add export account's private key

### LINKED ISSUE #277: Enable export privateKey for single accounts
Sometimes the user would need to access the privateKey of a single account for use outside the wallet.

Todo:
- [ ] Create a design flow for this
- [ ] Implement export account privateKey screen

LuizAsFight:
Closes #277


packages/app/src/systems/Vault/services/VaultServer.ts
```diff
@@ -159,6 +163,12 @@ export class VaultServer extends EventEmitter {
     const vault = await this.manager.exportVault(vaultId);
     return vault.secret || '';
   }
+
+  async exportPrivateKey({
+    address,
+  }: VaultInputs['exportPrivateKey']): Promise<string> {
+    return this.manager.exportPrivateKey(Address.fromString(address));
```
luizstacio, line 170: We should ask for the password before exporting the privatekey

luizstacio (changes requested):

luizstacio:
Editing and exporting could be moved to a menu instead of being on the item. This would open space to `hide account` also.

packages/app/src/systems/Account/components/AccountList/AccountList.tsx
```diff
@@ -8,14 +8,16 @@ import { AccountItem } from '../AccountItem';
 export type AccountListProps = {
   accounts?: Account[];
   isLoading?: boolean;
-  onPress?: (account: Account) => void;
+  onPress: (account: Account) => void;
```
eswarasai, line 11: @LuizAsFight thought we wanted `onPress` to be optional?
```suggestion
  onPress?: (account: Account) => void;
```

LuizAsFight: you're right

luizstacio (changes requested):
When typing a wrong password, the wallet should be locked, or we should show an error on the confirm password screen.

Currently, it's opening the export page and showing an error on the toast. 

Screenshot 2023-03-27 at 22 13 44

luizstacio (approved):

PR #621, opened 2023-03-13
feat: change create wallet flow

### LINKED ISSUE #102: Change create wallet flow, to achieve state full steps
On the current flow, when a user is creating a Wallet and the pipeline is broken for wherever reason (refresh, etc), the creation flow needs to be restarted. 

The possible improvement to do this is to re-order the flow;
1. Ask the user for a password
2. Show the seed phrase
  2.1 Store the encrypted data on storage
3. If the users refresh or anything happen
  3.1 Show the password unlock page
4. Show seed phrase
5. Finish flow

tomiiide:
## what does this PR do?
This PR reorgs the sign up flow to save the state of sign up, and let's the user resume a sign up in case of any issues.

## Demo
https://www.loom.com/share/2f53dab598924d50a7d22e8c436e5143

## Changes
1. New `ConfirmMnemonic` component
2.Password is now created first, before showing and confirming passphrase 
3. Confirm wallet screen uses a new UI. User arrange scrambled passphrase instead of copying and pasting on the next screen.
4. User can resume signup after password is created, if they reload the page or the extension crashes for any reason.

## Tasks
- [x] Create the `ConfirmMnemonic` component
- [x] Update the `signUpMachine` to align with the new flow.
- [x] Update the jest and  e2e tests for the machine
- [x] Update the `signUpMachine` to work with recoverWallet functionality
- [x] Update the `RecoverWallet` screen
- [x] Fix the `invalidMnemonic` state on the `createWallet` flow. 

Closes #102


packages/app/src/systems/SignUp/machines/signUpMachine.ts
```diff
@@ -43,13 +44,24 @@ type MachineServices = {
   setupVault: {
     data: Maybe<Account>;
   };
+  saveSignUp: {
+    data: Maybe<FormValues>;
+  };
+  getSavedSignup: {
```
eswarasai, line 50:
```suggestion
  getSavedSignUp: {
```

packages/app/src/systems/SignUp/machines/signUpMachine.ts
```diff
@@ -199,6 +240,13 @@ export const signUpMachine = createMachine(
       assignAccount: assign({
         account: (_, ev) => ev.data,
       }),
+      assignSavedData: assign({
+        data: (_, ev) => ({
+          mnemonic: (ev as MachineServices['getSavedSignup']).data.mnemonic,
```
eswarasai, line 245:
```suggestion
          mnemonic: ev.data.mnemonic,
```

packages/app/playwright/e2e/RecoverWallet.test.ts
```diff
@@ -19,47 +16,10 @@ test.describe('RecoverWallet', () => {
     page = await browser.newPage();
   });
 
-  test('should be able to recover a wallet', async () => {
```
matt-user, line 22: Why was this test deleted?

packages/app/src/systems/Core/components/Mnemonic/Mnemonic.tsx
```diff
@@ -49,11 +49,6 @@ export function Mnemonic({
     fillArray(initialValue, format)
   );
 
-  async function handleCopy() {
```
matt-user, line 52: Should we still allow a user to copy their secret phrase or are we basically forcing them to write it down physically somewhere?

packages/app/src/systems/SignUp/components/MnemonicConfirm/MnemonicConfirm.tsx
```diff
@@ -0,0 +1,247 @@
...
+};
+
+export function MnemonicConfirm({
+  words = [],
+  onFilled,
+  onNext,
+  onCancel,
+  positions,
+  readOnly,
+  defaultValue = fillArray([], 9),
```
matt-user, line 43:
```suggestion
  defaultValue = fillArray([], NUM_WORDS_TO_CONFIRM),
```
this isn't necessary but we could add
```
export const NUM_WORDS_TO_CONFIRM = 9
```
to the top of this file and then import and use the const in the `MnemonicConfirm` test.  That way if we ever change this number the tests won't break

LuizAsFight:
when copying seed phrase manually, we get this result:
```
addict
worth
grace
zone
beauty
empower
bacon
neck
seven
chef
wrong
disease
fall
burger
file
shallow
umbrella
rival
romance
turkey
mirror
wash
exercise
diesel
```

image



### Suggestion:
we don't wanna incentive copying, but if user wants to copy manually would be nice to copy in the correct format.  
maybe chaging html structure we can achieve :
```
addict worth grace zone beauty empower bacon neck seven chef wrong disease fall burger file shallow umbrella rival romance turkey mirror wash exercise diesel
```

LuizAsFight:
in the "Confirm Recovery Phrase" screen, it's showing words number in a random order, and it's not showing 24 words to select.

image

LuizAsFight:
I thought it was a bug at first, because last screen has all 24 words in order, then this screen has only 12 words, and not in order.

image


### Suggestion
to improve ux we should:
- confirm 1 word per time (randomly)... do it 3 times
- when confirming
  - ask for example: Type your 22th word. then show 5 options for the user to choose. 1 of them is the correct one, other 4 can be random (doesn't need to be contained in current 24 seed phrase)


doing this we will achieve better experience using single form approach

LuizAsFight:
1- I filled my seed words to confirm, then clicked ok... then got stuck in infinite loading

after refreshing the page I get this image error:

image

LuizAsFight:
in this page I can be clicking in the checkbox-next forever

https://user-images.githubusercontent.com/8636507/229001913-f1fd6a73-bac6-441f-b3d9-ee5be1cb1349.mov

LuizAsFight:
first screen is create wallet screen
image

in "Create Wallet", I need to put password first
in "Import Wallet", I need to put seed phrase first
we will also have "add hardware wallet case", which basically doesnt need password to encrypt anything (as we dont have access to private key). , but still good to have password to be able to lock/unlock , and also later the user can add other accounts that comes from private key or even new accounts.

to support all cases and keep a consistent UX we should:
- put create password screen as first screen
  - include a nice "Welcome" text to don't be a dry "create password" screen. exampe: Welcome to Fuel Wallet. Please create your password to start using Fuel Wallet.

LuizAsFight:
after importing from an existing seed phrase, then filling the password, I get locked in "loading" state forever
image

LuizAsFight:
> refreshing the page to try again, when trying to put password in "create wallet" screen I'm also getting infinity loading
> 
> maybe it's related/only happen after prev errors. I will try cleaning everything and start from scratch, but this error should be explored
> 
> image

I tried  "clear site data" then refresh page and try again... still stuck when putting password
image


then I removed the extension, and added it again. then it worked

LuizAsFight:
first I select I saved my seed phrase, then I go to confirm screen. 
in confirm screen, if I click "close", I go back to first step, and need to create password again.


when clicking "close" it should go again to the "backup seed phrase" screen...
after changing password to be the first screen, this makes even more sense


image

image

LuizAsFight:
after cancelling in confirm seed phrase step(like mentioned in prev comment), then creating again.

when trying to confirm seed phrase again, it says my words are wrong, but they're not(I'm sure). maybe it's some confusion between 2(or more) different seed phrases created? 

image

LuizAsFight:
if user put words incorrectly, it's end of line... user can only:

1- close and go to start again
2- refresh the page to confirm from scratch. 

image



in the new "single form" flow, if user mistakes one of 3 confirmations, we give the error feedback and show a button to redirect user to second screen ("Create a new Fuel Wallet"). then if user clicks on "Create Wallet", seed phrase generated needs to be different from prev one (as user couldn't confirm).

user will be in this loop until he can correctly save his seed phrase and confirm it for us

LuizAsFight (changes requested):
changes requested in prev comments. didn't review code yet, only functional

LuizAsFight:
steps to reproduce:
- click on "Create Wallet"
- fill a password
- click on next
- refresh the page
- do same process again.

image

LuizAsFight:
> 

this one wasn't made @tomiiide

LuizAsFight (changes requested):
I cannot change to use other size of mnemonic phrase anymore in Import Wallet  screen

image

LuizAsFight (changes requested):
1- Click in Create Wallet - reach the Backup your Recovery Phrase screen
2- then refresh the page, or click back to start again
3- Click in Create Wallet - react the Backup your Recovery Phrase screen again
4- phrase will be different than first one
5- now go to next page (Confirm Phrase)
6- first time opened, will show phrase after refresh. But if you refresh the page, will show phrase before refresh. Which will be the accepted one in the flow.

If a new phrase is generated, the phrase generated before refreshing should never be shown again

https://user-images.githubusercontent.com/8636507/234130728-4161b77f-b4d5-4343-96aa-f2ef8011a990.mov

packages/app/src/systems/SignUp/components/MnemonicWrite/MnemonicWrite.test.tsx
```diff
@@ -48,11 +18,12 @@ describe('MnemonicWrite', () => {
...
-    const btnPaste = screen.getByText('Paste');
-
-    act(() => {
-      btnPaste.click();
+    const buttons = screen.getAllByRole('button');
+    // remove the last two buttons
+    buttons.slice(0, buttons.length - 2).forEach((button) => {
+      act(() => {
+        button.click();
+      });
```
luizstacio, line 26: Can we add a aria-label to the button instead?

packages/app/src/systems/Vault/services/utils.ts
```diff
@@ -0,0 +1,25 @@
+export function shuffle<T extends string | number>(array: T[]): T[] {
```
luizstacio, line 1: I think we can move this under `Core/utils/mnemonic`?

packages/app/src/systems/Vault/services/VaultServer.ts
```diff
@@ -160,11 +177,35 @@ export class VaultServer extends EventEmitter {
...
+  }
+
+  async confirmMnemonic({
+    words,
+  }: VaultInputs['confirmMnemonic']): Promise<boolean> {
+    const mnemonic = (await this.exportVault({ vaultId: 0 })).split(' ');
+    const isValid = words.every((word, index) => {
+      return word === mnemonic[index];
+    });
+    return isValid;
```
luizstacio, line 206: I think here you can create a HDWallet using the words provided and do the match by first address this will remove the need to have the vault unlocked.

packages/app/src/systems/Vault/services/VaultServer.ts
```diff
@@ -160,11 +177,35 @@ export class VaultServer extends EventEmitter {
     vaultId,
     password,
   }: VaultInputs['exportVault']): Promise<string> {
-    await this.manager.unlock(password);
+    if (password) await this.manager.unlock(password);
```
luizstacio, line 180: This is a security rule to avoid the export been call without a password even if the vault is unlocked. This if needs to be removed.

packages/app/src/systems/SignUp/components/MnemonicConfirm/MnemonicConfirm.tsx
```diff
@@ -0,0 +1,230 @@
...
+        src={relativeUrl('/signup-illustration-1.svg')}
+        width={129}
+        height={116}
+        alt="Signup Illustration"
+      />
+      <Header
+        title="Confirm your Recovery Phrase"
+        subtitle="Fill in the missing words to confirm your recovery phrase."
+      />
+      <Stack gap="$3" css={{ width: 400 }}>
```
luizstacio, line 91: Move all css props to styles.

packages/app/src/systems/SignUp/machines/signUpMachine.ts
```diff
@@ -28,6 +28,8 @@ export enum SignUpType {
 type FormValues = {
   mnemonic?: string[];
   password?: string;
+  mnemonicConfirmation?: string[];
+  positionsForConfirmation?: number[];
 };
```
luizstacio, line 33: We should not have the mnemonic and password on the context. Can we remove it? Any technical blockers here?

packages/app/src/systems/SignUp/pages/WelcomeScreen/WelcomeScreen.tsx
```diff
@@ -1,12 +1,21 @@
...
 
 export function WelcomeScreen() {
   const navigate = useNavigate();
+  const isSigningUp = useIsSigningUp();
+
+  useEffect(() => {
+    if (isSigningUp) {
+      navigate(Pages.signUpCreateWallet());
+    }
+  }, [isSigningUp]);
```
luizstacio, line 18: Why we need this effect here could we do this by using a Route Guard?

luizstacio (changes requested):

packages/app/src/systems/SignUp/pages/WelcomeScreen/WelcomeScreen.tsx
LuizAsFight: would be a big change but I think this whole signup flow should be done using machine instead of route/effect/navigate mixed with machines in some points.

packages/app/src/systems/SignUp/machines/signUpMachine.ts
tomiiide: we need the `mnemonic` on the state for the MnemonicConfirm component, on the first flow before a reload case.

but we can remove the password.

tomiiide: @luizstacio 

To create the Vault, we need two inputs `mnemonic` and `password`, that are set at different steps / components.

I tried removing the `password` from the context.

It works for the create wallet flow because the when user sets the password on the first step, we generate the mnemonic, and we can create the vault since we have both inputs in one step.

but it doesn't work for the recovery flow, because the Vault is created at the end of the flow, and we would need the password, and the user entered mnemonic, we need to store the password in the context while the user enters the mnemonic.

luizstacio:
Due to recent changes on the signUp flow in #625 this PR will need to be re-write, and should be reprioritize.

PR #624, opened 2023-03-20
refactor: remove fuelhat to simplify node infra

luizstacio:
This PR removes the wrapper `fuelhat`. The idea is to remove unnecessary complexity on the codebase.

Today the `fuelhat` is not handling errors well. If you try to run it without having docker running or some other errors related to docker execution happens, the CLI not show erros and just logs as it had work. This could be solved by fixing the `fuelhat` code, but this is the unecessary complexity it adds in this case where just targeting directly docker compose with alias scripts on package.json will solve all the errors, and no future maintenance is needed.


docker/Makefile
```diff
@@ -0,0 +1,9 @@
+start:
+	docker compose -p dev --env-file .env up -d --build
```
LuizAsFight, line 2:
```suggestion
up:
	docker compose -p dev --env-file .env up -d --build 
```

docker/Makefile
```diff
@@ -0,0 +1,9 @@
+start:
+	docker compose -p dev --env-file .env up -d --build 
+
+start-test:
+	docker compose -p test --env-file .env.test up -d --build
+
+stop:
+	docker compose -p dev down --rmi local -v --remove-orphans
+	docker compose -p test down --rmi local -v --remove-orphans
```
LuizAsFight, line 9:
```suggestion
clean:
	docker compose -p dev down --rmi local -v --remove-orphans
	docker compose -p test down --rmi local -v --remove-orphans
```

LuizAsFight: can we create 2 new ones: ?

- `down` - will only get docker-compose down, but not clean everything.
- `down-test` - same thing of above, but for test docker


reason for this is that not always we want to clean the docker... sometimes we want to stop the node, but keep contracts and other transaction we might already have inside local node

package.json
```diff
@@ -43,10 +43,11 @@
...
     "lint:fix": "pnpm lint:check --fix",
-    "node:dev": "fuelhat node",
-    "nodes:start": "fuelhat node start -d && fuelhat node start -t -d",
-    "nodes:stop": "fuelhat node stop -d && fuelhat node stop -t -d",
-    "nodes:reset": "pnpm nodes:stop && pnpm nodes:start",
+    "node:dev": "make -C ./docker start",
+    "node:test": "make -C ./docker start-test",
+    "nodes:start": "run-s node:dev node:test",
+    "nodes:stop": "make -C ./docker stop",
+    "nodes:reset": "run-s nodes:stop nodes:start",
```
LuizAsFight, line 50: can we change to:

`node:up` 
`node:test:up` 
`node:down`
`node:test:down`
`node:clean`
`node:test:clean`


`nodes:up`
`nodes:down`
`nodes:clean`
`nodes:reset`

LuizAsFight (changes requested):

LuizAsFight (approved):
looking good !

PR #637, opened 2023-03-27
Update FuelLoaded.tsx

Haywoodsa:
// Fuel loaded handler
const onFuelLoaded = () => {
  setFuel(window.fuel);
};

// If fuel is already loaded, call the handler
if (window.fuel) {
  onFuelLoaded();
}

// Listen for the fuelLoaded event
document.addEventListener("FuelLoaded", onFuelLoaded);

### DIFF
--- a/packages/docs/examples/FuelLoaded.tsx
+++ b/packages/docs/examples/FuelLoaded.tsx
@@ -42,3 +42,16 @@ export function FuelLoaded() {
     </ExampleBox>
   );
 }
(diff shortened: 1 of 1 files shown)


Haywoodsa:
const isConnected = await fuel.connect();
console.debug("Connection response", isConnected);

luizstacio:
@Haywoodsa can you explain better the motivation for the changes? It doesn't look to be correct.

Issue #632, opened 2023-03-23
Add account name on home page and transaction view


luizstacio:
Currently, the account name is only visible on the account list. This task should create a component that gets the account's name from the account list, and if not found, it shows an unknown label.

<img width="743" alt="Screenshot 2023-03-23 at 10 52 00" src="https://user-images.githubusercontent.com/3941923/227225470-8767fb3a-9b71-4cb8-b2e5-65c531435380.png">
<img width="524" alt="Screenshot 2023-03-23 at 10 53 46" src="https://user-images.githubusercontent.com/3941923/227225496-e0dfbe27-2b89-4d25-98ae-6505b8c57c93.png">

LuizAsFight:
left is design, right is current PR implementation

notice that: 
1- "Account 1" font size is bigger in right.
2- "fuel...asd" font size is bigger in right
3- "ETH 0.00" font size is smaller in right
4- spacing to the bottom after "ETH 0.00" is smaller in right
5- eye icon is a bit smaller in right


<img width="718" alt="image" src="https://user-images.githubusercontent.com/8636507/229011327-67965607-3519-4ace-9ba4-df7cfc753a37.png">

PR #641, opened 2023-03-28
chore: added account name in balance widget

eswarasai:
Fixes: #632


LuizAsFight (changes requested):
e2e test broken + conflicts

LuizAsFight:
heres design link: https://www.figma.com/file/uEckQY4EaQ5rXwyISg6d1g/Fuel-Wallet?node-id=10-7410&t=679nML87sZdgZmPf-4

packages/app/src/systems/Account/components/BalanceWidget/BalanceWidget.tsx
```diff
@@ -144,4 +155,7 @@ const styles = {
     padding: '0 3px !important',
     borderRadius: 8,
   }),
+  name: cssObj({
+    margin: '0px 0px -5px',
```
LuizAsFight, line 159: do we need this negative margin? if we could do in other way would be nice

packages/app/src/systems/Transaction/components/TxRecipientCard/TxRecipientCard.tsx
```diff
@@ -23,12 +23,15 @@ export const TxRecipientCard: TxRecipientCardComponent = ({
...
 }) => {
+  const { accounts } = useAccounts();
   const address = recipient?.address || '';
   const isValidAddress = isB256(address) || isBech32(address);
   const fuelAddress = isValidAddress
     ? Address.fromString(address).toString()
     : '';
   const isContract = recipient?.type === AddressType.contract;
+  const name =
+    accounts?.find((a) => a.address === fuelAddress)?.name || 'unknown';
```
LuizAsFight, line 34: I think `unknown` it's not looking good on this context

image
image

We should move address to first position (avoid holes in layout), and put the account name below the address. also reduce highlight of account name, reducing font-size and making spacing smaller

LuizAsFight (changes requested):

packages/app/src/systems/Transaction/components/TxRecipientCard/TxRecipientCard.tsx
eswarasai: @LuizAsFight - Switching the positions of Account `name` and `address` breaks design consistency. I'd probably recommend adding a warning icon ⚠️ beside the `unknown` to emphasise the user to double check the address they're transferring funds to, thoughts?
cc: @luizstacio

luizstacio: Let's keep the consistency and the name on the top, and use unknow when we don't have it. I think is no need for the warning icon at this point.

luizstacio (changes requested):
I think the layout for the balance widget needs some more vertical space in the new design. The height was increased a little bit. Also, we can remove the margin-top of the container.

eswarasai:
> we can remove the margin-top of the container

@luizstacio - Reduced margin-top to have the account name center aligned with that of the hidden icon on the right side

LuizAsFight:
@luizstacio `unknown` text is weird and should be removed in my opinion but I won't block the PR because of this, you guys decide it. I never seem unknown when approving transaction in metamask for example.

the only data **_needed_** to a transfer is the address.. main data should be shown first, extra data after.


other text than "unknown" will be shown only when the account is registered in the wallet, which means the user owns it and have the keys. every transfer to someone else will have this "unknown" test, which brings unsafe thoughts in user's mind

packages/app/src/systems/Transaction/components/TxRecipientCard/TxRecipientCard.tsx
LuizAsFight: `unknown` is weird and should be removed in my opinion but I won't block the PR because of this, you guys decide it. I never seem `unknown` when approving transaction in metamask for example.

the only data _**needed**_ to a transfer is the address
account name is extra data, to increase level of information provided. but it's not needed.
regarding UX highlighting order, main data should be shown first, extra data after.

this screen will **almost always** show "unknown", which brings unsafe thoughts in user's mind.
it will show other thing only when the account is registered in the wallet, which means the user owns it and have the keys

LuizAsFight: cc @luizstacio

luizstacio (approved):

packages/app/src/systems/Transaction/components/TxRecipientCard/TxRecipientCard.tsx
LuizAsFight: then when it's not unknown, we show...

LuizAsFight (approved):

Issue #630, opened 2023-03-23
Add hide account button on UI


eswarasai:
Implement hide account functionality in the Accounts list screen

luizstacio:
@eswarasai PR #619 will be adding a menu on the account item, where the button to hide the account should go.

PR #674, opened 2023-04-07
feat(app): add hide account to dropdown

eswarasai:
Fixes: #630


luizstacio (changes requested):
Just missing E2E tests

luizstacio (changes requested):
I think we can add some more filters;
- Hide button should only be visible if the user has multiple accounts. If the user has only one visual history, it should not show the hidden button.
- It should not be possible to select a hidden account.
Feedbacks
- After hiding/unhiding, the account list should be refreshed.
Issues
- After adding a new account, the list is not refreshing anymore and the account does not appear on the list.

packages/app/src/systems/Account/components/AccountItem/AccountItem.tsx
```diff
@@ -75,6 +77,15 @@ export const AccountItem: AccountItemComponent = ({
           Export Private Key
         </Dropdown.MenuItem>
       ),
+      onToggleHidden && (
+        <Dropdown.MenuItem
+          key="hide"
```
matt-user, line 82: should this key contain some info unique to the account?  Without unique account info wouldn't we have duplicate AccountItem components with the same 'hide' key?

packages/app/src/systems/Account/components/AccountList/AccountList.tsx
```diff
@@ -19,17 +20,20 @@ export function AccountList({
...
+  const [isPrimaryAccount, setIsPrimaryAccount] = useState(false);
 
   function toggle() {
     setShowHidden((s) => !s);
   }
 
   useEffect(() => {
-    const hiddenAccounts = (accounts ?? []).some((acc) => acc.isHidden);
-    setAnyHiddenAccounts(hiddenAccounts);
+    const hiddenAccounts = (accounts ?? []).filter((acc) => acc.isHidden);
```
matt-user, line 34: I see that there are a couple of `accounts ?? []` in this component.  Imo it would be better to have some variable 
`const accounts = accounts ?? []` and then use that variable throughout the component.

matt-user: Also what are the cases for when accounts is undefined?  Would it be possible to change the type of the accounts prop to not accept undefined?

packages/app/src/systems/Account/components/AccountList/AccountList.tsx
```diff
@@ -19,17 +20,20 @@ export function AccountList({
...
+    setAnyHiddenAccounts(hiddenAccounts.length > 0);
+    setIsPrimaryAccount((accounts ?? []).length - hiddenAccounts.length === 1);
```
LuizAsFight, line 36: both hiddenAccounts and PrimaryAccount logics could be implemented in `useAccounts` hook

I would create selectors like:
```
  hasHiddenAccounts(state: AccountsMachineState) {
    return (state.context?.accounts?.some((acc) => acc.isHidden));
  },
  canHideAccounts(state: AccountsMachineState) {
    return (state.context?.accounts?.filter((acc) => !acc.isHidden)?.length || 0) > 1;
  },
```

and export data from the hook... then can remove this whole `useEffect` `useState` `setAnyHiddenAccounts` ... logic

packages/app/src/systems/Account/components/AccountItem/AccountItem.tsx
eswarasai: Dropdown menu is visible only when the user clicks on the 3 dots icon of an account. So it's being rendered only once on the screen, hence no duplicates.

LuizAsFight (changes requested):

packages/app/src/systems/Account/components/AccountList/AccountList.tsx
eswarasai: Thanks for pointing this out 🙏

packages/app/src/systems/Account/components/AccountList/AccountList.tsx
eswarasai: > Also what are the cases for when accounts is undefined? Would it be possible to change the type of the accounts prop to not accept undefined?

We're getting `accounts` from the state machine, where the initial context is set to `undefined` instead of an empty array. @LuizAsFight @luizstacio can probably recommend if we'd like to update the default state of the `accountsMachine` to be an empty array instead of `undefined`

LuizAsFight: this is fixed I guess? not needed `accounts ?? []` anymore as data is coming from  `useAccounts` hook

matt-user (approved):

luizstacio (approved):

Issue #453, opened 2023-01-20
Docs: Open all the sub menus


tomiiide:
### Motivation

Currently, the sub-items of the docs, are closed by default, which makes it difficult for users to explore, and find the sections.

The idea is to bring it open always, and add the possibility to close it if the user wants.

### Possible implementations

Set the `packages/docs/src/components/SidebarSubmenu.tsx`  component `isActive` state value to be `true` by default.\
https://github.com/FuelLabs/fuels-wallet/blob/eb1b1c4e6f4aa058a93e9772c75155813502f636/packages/docs/src/components/SidebarSubmenu.tsx#L18

Issue #462, opened 2023-01-23
Integrate your DApp section on docs


luizstacio:
This section replaces the old two sections, `how to use` and `TypeScript`.

- Introduction
    - Integrate your DApp seamlessly with Fuel Wallet using our simple API, which is injected directly onto the page window. This eliminates the need for additional dependencies, allowing you to perform key actions effortlessly. For even more powerful and customizable integration, use Fuel Wallet in conjunction with the Fuel TS SDK.
- How to use (same content as previous)
- TypeScript Support (same content as previous)

luizstacio:
@sarahschwartz, maybe with some changes your PR  #676 can solve this issue?

PR #676, opened 2023-04-07
docs: move dev docs to "For Developers" section, add "How to Use" page for non-devs

sarahschwartz:
This PR:
- moves the developer-focused docs into a "For Developers" section
- renames the old "How to Use" page to "Getting Started"
- renames the old "API" page to "Reference"
- adds a new page called "How to Use" that is user-focused and explains how to use the wallet extension
- small edits to the install page wording

close: #453 #462


packages/docs/docs/how-to-use.mdx
```diff
@@ -2,135 +2,226 @@
...
+    src="/wallet/import-account-screen.png"
+    height="500"
+  />
+</div>
+
+### Edit an account
+
+To edit an account, click on the horizontal ellipsis to the right of the account and select "Edit".
+
+<div style={{ display: 'flex', gap: '20px', flexWrap: 'wrap' }}>
```
luizstacio, line 41: We should create a component for this to avoid code repetition. Also we can add an automatic scroll on the y-axis to prevent images from being broken in lines.

luizstacio (changes requested):

luizstacio (changes requested):
Also, can we make the menus open by default on this PR?

tomiiide:
Great work @sarahschwartz 

Just one suggested improvement from me.
I think video / GIF demos would be better than screenshots.

LuizAsFight (dismissed):

luizstacio (changes requested):
I think we can do some improvements here;
1. On the container with scroll add the scroll in auto to show only if the content overflow and only on the x-axis.
2. Start the videos on pause and wait for the reader to click on play, this will avoid the videos to load until the user do a play.

luizstacio (approved):
Thanks for the work here 💚

Issue #500, opened 2023-01-24
Change Wallet Password


luizstacio:
[FRO-70](https://linear.app/fuel-network/issue/FRO-70/change-wallet-password)

User Story:
As an engineer, I shall refactor the Change Password feature to improve error handling, positioning of elements, and user experience.

Acceptance Criteria:

Display an error message below the input fields when the new password and confirm password inputs do not match the rules.
Move the warning box to the bottom of the page.
If the submitted current password is incorrect, display an error message below the current password input field.
When the Change Password screen is opened, set the focus on the current password input field.
Display the "passwords don't match" error message only after the user leaves the input field (on blur event).

PR #691, opened 2023-04-13
fix: change password UI/UX issues

eswarasai:
Fixes: #500

### DIFF
--- a/packages/app/src/systems/Settings/pages/ChangePassword/ChangePassword.tsx
+++ b/packages/app/src/systems/Settings/pages/ChangePassword/ChangePassword.tsx
@@ -1,5 +1,5 @@
 import { cssObj } from '@fuel-ui/css';
-import { Alert, Button, Flex, InputPassword } from '@fuel-ui/react';
+import { Alert, Button, Flex, Focus, InputPassword } from '@fuel-ui/react';
 import { yupResolver } from '@hookform/resolvers/yup';
 import { useForm } from 'react-hook-form';
 import { useNavigate } from 'react-router-dom';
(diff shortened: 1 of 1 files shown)


luizstacio (changes requested):
- When the input has an error, and the user starts to change the input it should clean the error.

luizstacio:
> 

Once the user types the password again and now is correct, it shouldn't have an error, but currently, the user needs to leave the input to see that it is correct. Cleaning the error once start to type again it solves this issue.

luizstacio:
> > > 
> > 
> > 
> > Once the user types the password again and now is correct, it shouldn't have an error, but currently, the user needs to leave the input to see that it is correct. Cleaning the error once start to type again it solves this issue.
> 
> on this way the user need to blur again to know if error is really fixed why don't just validate on `change` instead of `blur` then?

Avoid the error to be shown as the user type.

LuizAsFight:
> Avoid the error to be shown as the user type.

@luizstacio then what we need is don't show the error straight when typing in first interaction. 
1- fiels are empty, user start typing. At this point we will show any error only after user blur the "Confirm Password" field
2- user finished fill "Confirm Password" field, and blurred.
3- passwords don't match, now errors are showing
4- after this point, we check on "change"

does that make sense?


also other option is splitting this in 2 steps (trust wallet app does in this way if you wanna check and test)
complexity would be less and more isolated space to show error
1. User type the "Password" field
    - popup password strength gets shown and validate the pass
2. User clicks in "Next"
3. User type the "Confirm Password" field
    - at this point could use logic mentioned above also. first check only on blur, then on change if has errors.
    - show "Back" button to move back and change password in step 1 again
    - show "Confirm" button that is disabled until passwords are matching

luizstacio (changes requested):
The strength popup is not closing when moving to the second password field. Looks related to this PR as on master it's working correctly. Maybe because we are using the onBlur?

LuizAsFight (approved):

packages/app/src/systems/Settings/pages/ChangePassword/ChangePassword.tsx
```diff
@@ -54,8 +56,19 @@ export function ChangePassword() {
     setError,
     setValue,
     formState: { isValid },
+    trigger,
   } = form;
 
+  const debouncedValidate = debounce(() => {
+    trigger('confirmPassword');
+  }, 500);
```
luizstacio, line 64: This looks a bit concerning as it will be render on every render, possibly creating memory leaks would be good to have this inside a useCallback without dependencies that would enable to clean the debounce on each render.

```suggestion
  const debouncedValidate = useCallback(debounce(() => {
    trigger('confirmPassword');
  }, 500), []);
```

luizstacio, line 64: This looks a bit concerning as it will be render on every render, possibly creating memory leaks would be good to have this inside a useCallback without dependencies that would enable to clean the debounce on each render.

```suggestion
  const debouncedValidate = useCallback(debounce(() => {
    trigger('confirmPassword');
  }, 500), []);
```

luizstacio (changes requested):

luizstacio (approved):

Issue #610, opened 2023-03-03
Change network add workflow


luizstacio:
[FRO-55](https://linear.app/fuel-network/issue/FRO-55/change-network-add-workflow)

On Create Flow;
- Split add network URL in two steps;
- When the user enters the screen, just the network URL should be visible
- After filling the URL, the screen should load the name from the Network URL and show the input name already filled
   - If the fetch fail, we should show a warning for the user but not block the adding process.
- The user should be able to customize the name

On Edit Flow;
- The user should be able to edit both, and we don't need to fetch the name network URL again.

PR #723, opened 2023-04-24
feat(app): modify network add ux

eswarasai:
Fixes: #610 

### Notes:
- [x] Modified url regex validation instead of relying on the `new URL(url)` as it's returning `true` for `https://graphql` even though it really isn't valid provider url

### TODO:
- [x] Good to have debounce on the url validation and fetching name of the network, if url is valid


luizstacio (changes requested):
On the issue, we specify splitting the flow into two steps.
- First, fill in the URL, and after it's fetched, show the Network Name
- About adding the debounce, we can only fetch once the URL is valid this would enable to not show toast every keypress
- Also we can split the screen into steps where the button on the first state with only URL showing is next and after we have both we show save.

LuizAsFight (changes requested):
when creating new network, if type incorrect URL, I'm getting this toast error

image

We should show a better error in input URL field: "Invalid Network"

LuizAsFight (changes requested):
After typing correct URL, I don't get any sign that name is being loaded, and it can take time if nodes are slow.

We should show a loading feedback in name field when it's getting loaded

LuizAsFight (changes requested):
Even typing wrong URL, I'm able to save and move on with incorrect network. We should block saving if it's invalid url

LuizAsFight:
something like this would be cool to indicate it's a valid network @luizstacio  

image

eswarasai:
> we specify splitting the flow into two steps

@luizstacio - Reason for not splitting and instead swapping the positions of URL and Name is because we're using the same UI for Update Network flow. Splitting into 2 steps seems unnecessary, especially when we're not forcing the user to stick to the default Network Name. Felipe's feedback above would definitely improve the user experience for adding Network while keeping the flow simple. Let me know if that makes sense, thanks!

cc: @LuizAsFight

LuizAsFight:
add network flow in 2 steps is a good idea:

1. ask the Network URL only
    - ADD button is disabled
    - when we detect a valid URL
      - enable loading mode (disable field and show loading inside field)

2. once network data is loaded, show a "review" screen to confirm adding the network
    - don't allow to change the name at this point.
    - if click in "Change URL", go to prev step

edit network flow:
- can edit only name
- network URL field disabled


image
image



cc @luizstacio

luizstacio:
> add network flow in 2 steps is a good idea:
> 
> 1. ask the Network URL only
>    
>    * ADD button is disabled
>    * when we detect a valid URL
>      
>      * enable loading mode (disable field and show loading inside field)
> 2. once network data is loaded, show a "review" screen to confirm adding the network
>    
>    * don't allow to change the name at this point.
>    * if click in "Change URL", go to prev step
> 
> edit network flow:
> 
> * can edit only name
> * network URL field disabled
> 
> image image
> 
> cc @luizstacio

@eswarasai what you think about this? I think would make sense to do in this way and let the user change the name by editing and not on the time he is adding. This is a change on the spec but looks a good improvement. 

If this is too overkill we could just add the check as mentioned prior to ensure the URL is valid, and use the same space on the side to show the loading spinner when we are fetching the name of the network.

eswarasai:
> let the user change the name by editing and not on the time he is adding

@LuizAsFight - What if there's an error when trying to fetch the network name from chain info? For example, if I try to add [Beta 1](https://node-beta-1.fuel.network/graphql) network. Considering that the network name is mandatory and must be unique, if we don't have an edit option, it might block the flow?

Splitting the Add Network flow into 2 steps like you've mentioned above would definitely be an improvement to the user experience. Will update the PR 👍

LuizAsFight:
> > let the user change the name by editing and not on the time he is adding
> 
> @LuizAsFight - What if there's an error when trying to fetch the network name from chain info? For example, if I try to add [Beta 1](https://node-beta-1.fuel.network/graphql) network. Considering that the network name is mandatory and must be unique, if we don't have an edit option, it might block the flow?

if there's an error fetching the name, wallet will most probably don't support that network, right? then we can show the error above the input and block the insertion... same as if was an invalid url without `graphql` at the end? or maybe show `Network not supported`

ps.: we're improving but in the future we should have an even better way to check if we support that network before adding it to wallet

luizstacio (changes requested):
It should be possible to add urls that are not https: `http://localhost:4000/graphql`.

packages/app/.env.test
```diff
@@ -1,5 +1,5 @@
-VITE_FUEL_PROVIDER_URL=http://localhost:4001/graphql
-VITE_FUEL_FAUCET_URL=http://localhost:4041/
+VITE_FUEL_PROVIDER_URL=http://127.0.0.1:4001/graphql
+VITE_FUEL_FAUCET_URL=http://127.0.0.1:4041/
```
luizstacio, line 2: Why we need this change?

packages/app/src/systems/Network/pages/AddNetwork/AddNetwork.tsx
```diff
@@ -1,11 +1,42 @@
...
+    if (chainInfoError) {
+      form.setError('url', {
+        type: 'manual',
+        message:
+          chainInfoError === 'Network request failed'
+            ? 'Invalid network'
+            : 'Unsupported network',
+      });
+    }
+  }, [chainInfoError]);
```
luizstacio, line 39: We can move this logic to the react-form component it support url validations and also async validations.

luizstacio (changes requested):

packages/app/src/systems/Network/hooks/useNetworkForm.ts
```diff
@@ -15,9 +15,10 @@ function isValidUrl(url?: string) {
   if (!url) return false;
   // Note: new URL('https://graphql') returns `true`
   const pattern = new RegExp(
-    '^(https?:\\/\\/)?' +
+    '^(https?:\\/\\/)' +
```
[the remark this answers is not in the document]
eswarasai: > Since we removed the ? we require the urls to be prefixed by some sort of http: or https: prefix?

Yes, I think we should keep the protocol mandatory for the provider url

> Doesn't this pattern allow for urls like "https:\\"? is this the desired behavior?

`^(https?:\\/\\/)`=> Must start with `http` or `https`, `s?` means `s` is optional. `\\` is used to escape special characters when using `new RegExp()`. Regex will evaluate `https:\\` as false.
ref: https://stackoverflow.com/a/17863171

luizstacio (dismissed):

LuizAsFight (changes requested):
looking pretty good.

one minor comment:
if you notice `Connect` screen, you'll see that there's an animation whenever the step changes (also when opening the screen). can we copy that animation here?

LuizAsFight (changes requested):
animation is not consistent with connect screen one:

- first time input appears there is an animation, but when clicking "change url" to move back, animation don't happen again
- animation of second step is coming from left, should come from same direction

https://user-images.githubusercontent.com/8636507/236970349-8c5bdba2-27e5-4603-9de7-f5447f0b3783.mov

LuizAsFight (approved):
looking pretty good

Issue #382, opened 2022-12-20
Enable FuelWeb3 SDK to be use direct without the need to use the injected script


luizstacio:
[FRO-83](https://linear.app/fuel-network/issue/FRO-83/enable-fuelweb3-sdk-to-be-use-direct-without-the-need-to-use-the)

Today developers need to wait and use FuelWeb3 to be injected into the window. We should enable developers to create their on instances of FuelWeb3. For this, we should also add checks to enable developers to check if Wallet is installed. And other methods should return an error if not installed.

```ts
import { FuelWeb3 } from '@fuel-wallet/sdk';
const fuelWeb3 = new FuelWeb3();


await fuelWeb3.hasWallet(); // return true or false if installed or not
await fuelWeb3.connect(); // Throw error if not installed.
```

PR #729, opened 2023-05-02
feat: add multiple wallet connectors

luizstacio:
## Description

This pull request enhance the Fuel SDK by allowing developers and users to choose their preferred Wallet provider. With this feature, other community Wallets such as Fuelet can seamlessly integrate with DApps in the Fuel ecosystem, using the same communication protocol and interfaces.

The implementation enables the following flow:
- The **Wallet Provider** can use the `createConnector` method and the `ContentProxyConnection` class, passing the `wallet provider configuration` with the name of the current **Wallet**.
- **DApps** can use the `addConnector`, `listConnectors`, and `selectConnector` methods to allow users to choose which Wallets to use. The `window.fuel.listConnectors()` can be used to check if there is more than one conector available.

The execution architecture works as follows:
- When `createConnector` is called inside a `Page Script`, the method checks whether the `window.fuel` is already available. If yes, the provider configuration is appended to the providers list instead of creating a new instance of `Fuel`.
- When a method is called from fuel (e.g., `window.fuel.connect()`), the messages used to communicate will be sent to the currently selected `target`.
- To change the `target`, the application will call the `selectConnector` method, which finds and updates the `current provider wallet` by name.
- if the `target` is not equal to the current configuration, the `ContentProxyConnection` will reject the message received.

Possible extensions for the multiple Wallet feature include implementing a UI that can be opened directly on the DApp to choose the wallet provider.

## In this PR

- Implement Fuel Wallet Connectors features
- Add docs for the Fuel Wallet Connectors
- Improve docs `CodeImport` with `commentBlock` that enables to import examples from a specific comment block, removing the need to import using start and end lines that are more prone to error.
- Change from `acorn` to `acorn-loose` to enable import of codes with better error tolerance. This was needed to enable acorn to ignore `impor type` as it's not a recognized node.
- Improve Fuel Wallet SDK tests by create a stack where the only mocked pice is the Background Service, creating a more realistic test.

close: #732


packages/sdk/src/Fuel.ts
```diff
@@ -14,6 +15,33 @@ const FuelWeb3Privates: {
...
+  selectWalletProvider(providerName: string): void {
+    const providerConfig = this.walletProviders.find(
+      (provider) => provider.name === providerName
+    );
+
+    if (!providerConfig) {
+      throw new Error(`Wallet provider ${providerName} not found`);
+    }
+
+    this.targetWallet = providerConfig.name;
```
mpoplavkov, line 42: NIT: for me this looks like accessing a private contract of the `WindowConnection` class. In the constructor of `Fuel` you call the `WindowConnection` super constructor and pass `providerConfig.name`, but here you directly set `providerConfig.name` to a `WindowConnection`'s field. This is a bit error prone, cause someone later might use `targetWallet` in `WindowConnection`'s in a different way, then he will also need to change the behavior here. Maybe it would be better to add a `setupTargetWallet` method in the `WindowConnection` and call it in both places.

I have a lack of knowledge in TS, not sure if that's possible. Just decided to share some thoughts :)

mpoplavkov (approved):

packages/sdk/src/Fuel.ts
```diff
@@ -1,3 +1,4 @@
+import type { WalletProvider } from '@fuel-wallet/types';
```
LuizAsFight, line 1: regarding provider name:
- it defines how the user will connect with the network. will the user connect to `fuel-beta-3` production node directly? or will use some other provider like `alchemy`?


for this feature I think a best name would be: `WalletConnector`

then we allow other apps to create their own `Connectors` extending from `WalletConnector`. and we also can expose our won `FuelWalletConnector`
image

LuizAsFight (changes requested):

packages/sdk/src/Fuel.ts
LuizAsFight: adding to this idea, I think we should introduce other layer ahead of the connectors.

then injectFuel we can call the `FuelClient` which can receive options like:

```
new FuelClient({
  connectors: [
    new FuelWalletConnector(),
    new FueletWalletConnector(),
    new WatheverWalletConnector(),
  ]
});
```

then the `WalletConnectors` can have functions like:
- `isConnected` - is connected through this connector
- `isReady` - is Wallet available in user browser
- others..

on this way we can add to top layer `FuelClient` methods like:
`getConnectedWallet()` - will return `WalletConnector` that was used to connect
`getPossibleWallets()` - will return `WalletConnector`s that are ready to use in user browser

then we can remove the concept of `targetWallet`, because client will handle it

luizstacio: This can make sense right now we don't have a good way to provide the SDK to work alone on the FE. We need to first close #382 to think about this approach.

We want to provide a more integrated experience where the developer don't need to install multiple things in order to get access to all wallets the users has installed.

LuizAsFight: yeah providing a `FuelClient` is good solution for solving  #382 .

idea is that client can be created by just importing it, don't need to wait anything in `window`
```
import { FuelClient } from "@fuels-wallet/sdk";

new FuelClient({
  connectors: [
    new FuelWalletConnector(),
    new FueletWalletConnector(),
    new WatheverWalletConnector(),
  ]
});
```

then the client itself will search for window dependencies like `window.fuel` and other possibles.

this also allow us to in the future to create a react specific `FuelProvider` informing the `client`, and build our own experience of connecting wallet, with a cool connect popup and great ux. also to provide specific `hooks` to interact with wallets

packages/docs/docs/dev/connectors.mdx
```diff
@@ -0,0 +1,80 @@
...
+Inside the `pageScript.ts` you can call the method `createConnector` instead of creating a new object on the window. This method will add you Connector to the current,
+`window.fuel` element.
+
+```ts
+import { createConnector } from '@fuel-wallet/sdk';
+
+createConnector({ name: 'Wallet Name' });
+```
+
+In case you are using our Commnucation SDK, [Conent Page Script](https://developer.chrome.com/docs/extensions/mv3/content_scripts/) import `ContentProxyConnection` and inform the name of Wallet Exntesion.
```
LuizAsFight, line 69:
```suggestion
In case you are using our Communication SDK, [Conent Page Script](https://developer.chrome.com/docs/extensions/mv3/content_scripts/) import `ContentProxyConnection` and inform the name of Wallet Exntesion.
```

packages/docs/examples/Connectors.tsx
```diff
@@ -0,0 +1,123 @@
...
+      fuel.off(fuel.events.currentConnector, onCurrentConnector);
+    };
+    /* eventCurrentConnector:end */
+  }, [fuel]);
+
+  useEffect(() => {
+    if (!fuel) return () => {};
+    if (!fuel.listConnectors) {
+      setCompatibilityError(
+        'Current verison of Fuel Wallet not support Connectors!'
```
LuizAsFight, line 35:
```suggestion
        'Current version of Fuel Wallet doesn't support Connectors!'
```

packages/docs/src/type.d.ts
```diff
@@ -0,0 +1,3 @@
+declare module 'acorn-loose' {
+  export function parse(input: string, options?: any): any;
```
LuizAsFight, line 2: add comment for any

packages/sdk/src/tests/__mock__/MockBackgroundSerivce.ts
```diff
@@ -0,0 +1,163 @@
+import type { CommunicationMessage } from '@fuel-wallet/types';
```
LuizAsFight, line 1: rename this file name to fix typo

LuizAsFight (changes requested):

packages/docs/examples/Connectors.tsx
```diff
@@ -0,0 +1,123 @@
...
+      fuel.off(fuel.events.currentConnector, onCurrentConnector);
+    };
+    /* eventCurrentConnector:end */
+  }, [fuel]);
+
+  useEffect(() => {
+    if (!fuel) return () => {};
+    if (!fuel.listConnectors) {
+      setCompatibilityError(
+        'Current version of Fuel Wallet doesn't support Connectors!'
```
sarahschwartz, line 35:
```suggestion
        'Current version of Fuel Wallet doesn&apos;t support Connectors!'
```

packages/docs/src/constants.ts
```diff
@@ -13,6 +13,7 @@ export const MENU_ORDER = [
   'For Developers/Networks',
   'For Developers/Signing a Message',
   'For Developers/Reference',
+  'For Developers/Wallet Connectors',
```
sarahschwartz, line 16: I think the Reference should go last

packages/docs/docs/dev/connectors.mdx
```diff
@@ -0,0 +1,80 @@
...
+To list all available connectors you can use the `listConnectors()` method.
+
+<CodeImport
+  file="../../examples/Connectors.tsx"
+  commentBlock="listConnectors"
+/>
+
+### Selecting connector
+
+For selecting a different Wallet Connector use the `selectConnector(connectorName: string)` method.
```
sarahschwartz, line 26:
```suggestion
To select a different Wallet Connector use the `selectConnector(connectorName: string)` method.
```

packages/docs/docs/dev/connectors.mdx
```diff
@@ -0,0 +1,80 @@
...
+<CodeImport
+  file="../../examples/Connectors.tsx"
+  commentBlock="selectConnector"
+/>
+
+## Events
+
+### CurrentConnector
+
+This event is triggered when the Current Connector is change. The callback function receives the `FuelWalletConnector` object.
```
sarahschwartz, line 37:
```suggestion
This event is triggered when the Current Connector is updated. The callback function receives the `FuelWalletConnector` object.
```

packages/docs/docs/dev/connectors.mdx
```diff
@@ -0,0 +1,80 @@
...
+This event is triggered when the Connectors list is updated. The callback function receives the list of connectors, `Array<FuelWalletConnector>`.
+
+<CodeImport
+  file="../../examples/Connectors.tsx"
+  commentBlock="eventConnectors"
+/>
+
+## For Wallet Developers
+
+Wallet Connector is used for Thirdy Party Wallet devleoped by the community to integrate with the current fuel instance on the window, enabling a better experience for
```
sarahschwartz, line 55:
```suggestion
Third-party wallets developed by the community can use the Wallet Connector to integrate with the current `fuel` instance on the window, enabling a better experience for
```

packages/docs/docs/dev/connectors.mdx
```diff
@@ -0,0 +1,80 @@
...
+developers and users.
+
+### Creating a Wallet Connector
+
+Inside the `pageScript.ts` you can call the method `createConnector` instead of creating a new object on the window. This method will add you Connector to the current,
```
sarahschwartz, line 60:
```suggestion
Inside `pageScript.ts` you can call the method `createConnector` instead of creating a new object on the window. This method will add your Connector to the current,
```

packages/docs/docs/dev/connectors.mdx
```diff
@@ -0,0 +1,80 @@
...
+`window.fuel` element.
+
+```ts
+import { createConnector } from '@fuel-wallet/sdk';
+
+createConnector({ name: 'Wallet Name' });
+```
+
+In case you are using our Communication SDK, [Conent Page Script](https://developer.chrome.com/docs/extensions/mv3/content_scripts/) import `ContentProxyConnection` and inform the name of Wallet Exntesion.
```
sarahschwartz, line 69:
```suggestion
If you are using our Communication SDK [Content Page Script](https://developer.chrome.com/docs/extensions/mv3/content_scripts/) import `ContentProxyConnection` and pass the name of Wallet Extension.
```

packages/docs/docs/dev/reference.mdx
```diff
@@ -96,15 +102,62 @@ the DApp will also switch to that network. Support for specific DApp networks wi
...
+To list all available connectors you can use the `listConnectors(): Array<FuelWalletConnector>` method.
+
+<CodeImport
+  file="../../../sdk/src/tests/FuelWalletConnectors.test.ts"
+  testCase="listConnectors"
+/>
+
+### Has connector
+
+For checking if a connector is on the connectors list use `hasConnector(connectorName: string): boolean` method.
```
sarahschwartz, line 125:
```suggestion
To check if a connector is on the connectors list use the `hasConnector(connectorName: string): boolean` method.
```

packages/docs/docs/dev/reference.mdx
```diff
@@ -96,15 +102,62 @@ the DApp will also switch to that network. Support for specific DApp networks wi
...
+
+<CodeImport
+  file="../../../sdk/src/tests/FuelWalletConnectors.test.ts"
+  testCase="hasConnector"
+/>
+
+### Selecting connector
+
+For selecting a different Wallet Connector use the `async selectConnector(connectorName: string): Promise<boolean>` method.
+This method will check if the desired connector is install if not detect in 1 second it throws.
```
sarahschwartz, line 135:
```suggestion
This method will check if the desired connector is installed. If it is not detected in 1 second, the method throws an error.
```

packages/docs/docs/dev/reference.mdx
```diff
@@ -96,15 +102,62 @@ the DApp will also switch to that network. Support for specific DApp networks wi
...
+This method not triggers erros if the target Connector is not install.
+
+<CodeImport
+  file="../../../sdk/src/tests/FuelWalletConnectors.test.ts"
+  testCase="addConnector"
+/>
+
+### Remove connector
+
+For removing a Wallet Connector use the `removeConnector(connectorName: string): boolean` method.
```
sarahschwartz, line 154:
```suggestion
To remove a Wallet Connector use the `removeConnector(connectorName: string): boolean` method.
```

packages/docs/docs/dev/reference.mdx
```diff
@@ -145,16 +196,29 @@ This event is only triggered if the application has access to the current accoun
...
 
 ### Network
 
 This event is triggered when the client connects to a different network. The callback function receives the `network` object.
 
+<CodeImport file="../../examples/events/Network.tsx" commentBlock="network" />
+
+### CurrentConnector
+
+This event is triggered when the Current Connector is change. The callback function receives the `FuelWalletConnector` object.
```
sarahschwartz, line 210:
```suggestion
This event is triggered when the Current Connector is updated. The callback function receives the `FuelWalletConnector` object.
```

packages/docs/examples/Connectors.tsx
luizstacio: I fixed the string here. But not add &apos; as I like to avoid it on strings and keep it only when the text is directly on HTML.

sarahschwartz (approved):

LuizAsFight (approved):

PR #731, opened 2023-05-02
fix: include images on assets

luizstacio:
Include images inside the package asset-list to enable the package to be published to npm with metadata and images.


LuizAsFight (changes requested):
code looks good.

missing changeset and tests broken

LuizAsFight (approved):

Issue #705, opened 2023-04-17
Transfer DAI from address A to address B


yuanmangguo:
[FRO-51](https://linear.app/fuel-network/issue/FRO-51/transfer-dai-from-address-a-to-address-b)

### What version of Wallet are you using?

Fuel Wallet 0.7.2

### What is happening?

Only ETH can be transferred, other assets cannot be transferred.

### steps to reproduce

Transfer DAI from address A to address B

### How it should behave?

Transfer DAI from address A to address B

### What os are you seeing the problem on?

Windows Desktop

### What browser?

Chrome

### What is your browser version?

Version 112.0.5615.86 (Official Build) (64-bit)

luizstacio:
@yuanmangguo, thanks for the report I have tested here in found the issue, you can track it on this issue.

PR #737, opened 2023-05-02
fix: transfer alt assets

luizstacio:
close: #705


packages/app/src/systems/Transaction/services/transaction.tsx
```diff
@@ -214,13 +214,24 @@ export class TxService {
...
     return request;
   }
 
   static async fundTransaction(input: TxInputs['fundTransaction']) {
-    const transactionRequest = await TxService.addResources(input);
+    const { minGasPrice } = await input.wallet.provider.getNodeInfo();
+    const transactionRequest = await TxService.addResources({
+      ...input,
+      gasFee: minGasPrice,
+    });
```
LuizAsFight, line 234: `gasFee` is different from `gasPrice`

packages/app/src/systems/Transaction/services/transaction.tsx
```diff
@@ -214,13 +214,24 @@ export class TxService {
       amount: isEth(coin) ? bn(coin.amount).add(gasFee) : bn(coin.amount),
     }));
 
+    if (!coins.find(isEth)) {
+      coins.push({
+        assetId: NativeAssetId,
+        amount: gasFee,
+      });
+    }
+
```
LuizAsFight, line 223: as I'm understanding this code is made to add UTXO input to pay gas fee, correct?
if so, questions:

1- shouldn't this be done by sdk by default?
2- is this `coins.push` gonna be reflected in coins outputs? I'm asking because it has no `to` address
3- I see you're passing `minGasPrice`, which can be 1 unit for example. if UTXO input is of let's say 2 units, but real gasFee os transaction is 10 units, then this transaction will raise an error

LuizAsFight (changes requested):

packages/app/src/systems/Transaction/services/transaction.tsx
luizstacio: This coins are the resources needed to the transaction. It mean you need to account for the output coins but also the gasFee coins, when sending a alt asset, the outputs will not include ETH. Than the logic is if the coins required to the TX not include ETH, than we add ETH to the coins with the min amount to pay to the minGasFee. Than the SDK will check if this is enough to pay the gasUsed if not it automatically adds the amount required.

This logics we are doing to create a transfer can be removed and improve on the SDK but for now I tried to fix directly on the Wallet.

LuizAsFight (approved):

PR #741, opened 2023-05-05
feat: Auto generate account name for new accounts.

### LINKED ISSUE #502: Add new account skip naming step
[FRO-72](https://linear.app/fuel-network/issue/FRO-72/add-new-account-skip-naming-step)

When adding a new account, we can skip the name phase and add the account with an auto-generated name and let the user rename it later.

- When creating on add New Account, auto-generates the new account using `Account ${index}` without opening the screen to input the name.
- When importing from Private Key we can pre-fill the input.

tomiiide:
closes #502

## What does this PR do?
When a user is adding an account, Instead of asking a user to enter a name, we skip that step , and auto generate the name.

### demo
https://www.loom.com/share/fc4902cd025e4384929049320b99063c


packages/app/src/systems/Account/hooks/useAccounts.tsx
```diff
@@ -79,6 +79,12 @@ export function useAccounts() {
     selectors.canHideAccounts
   );
 
+  function addAccount() {
+    store.send(Services.accounts, {
+      type: 'ADD_ACCOUNT',
+    });
+  }
```
LuizAsFight, line 86: accountMachine is a global machine, for sending events to a global machine it's better to use `events.tsx` file inside `Accounts` system.

packages/app/src/systems/Account/services/account.ts
```diff
@@ -175,6 +174,18 @@ export class AccountService {
...
+    const suffix = randomSuffix ? Math.floor(Math.random() * 100) : '';
+    const accounts = await AccountService.getAccounts();
+    const count = accounts.length;
+    const name = `Account ${count + 1}${suffix}`;
+    const accountNameExists = await this.checkAccountNameExists(name);
+    if (accountNameExists) {
+      return this.generateAccountName(true);
+    }
+    return name;
+  }
```
LuizAsFight, line 187: this can use `getUniqueString`

it will add (1), (2), (3) at the end if has other account with same name.
`Account 2 (1)`
instead of current "random suffix" -> `Account 2 44` for example

something along this lines \\/

```suggestion
  static async generateAccountName() {
    const accounts = await AccountService.getAccounts();
    const count = accounts.length;
    const desiredName = `Account ${count + 1}`;
    const allNames = accounts.map(({ name }) => name);
    const name = getUniqueString(desiredName, allNames);
  }
```

LuizAsFight (changes requested):
1- I think we should keep loading state only for the button here... no needs to put list on loading, as we're going directly to home right after the action.
2- we should show a toast `"Account X" created successfully "

https://user-images.githubusercontent.com/8636507/236905914-8ccd87dc-1881-4a3b-9467-07c50d49dd0f.mov

LuizAsFight:
> 1- I think we should keep loading state only for the button here... no needs to put list on loading, as we're going directly to home right after the action. 2- we should show a toast `"Account X" created successfully "
> 
>  Screen.Recording.2023-05-08.at.15.42.45.mov

changes requested /\\ are simple to do now.

ideally I would like to see this flow: ( @luizstacio please check what you think before @tomiiide develops it) maybe we can do it in other PR as well:

1A- user clicks "Add Account"
1B- user clicks "Import from Private"
  1B- user informs private key and click ok
2- both 1A and 1B will be redirect to this page:
- show loading phrase
- show account loading component
example:
image

- Show a success screen 
example: 
image

eswarasai:
@LuizAsFight - Wouldn't it make sense to ask user confirmation in a dialog before proceeding with creation of an account, similar to what we've for delete account currently?
cc: @luizstacio

luizstacio:
> > 1- I think we should keep loading state only for the button here... no needs to put list on loading, as we're going directly to home right after the action. 2- we should show a toast `"Account X" created successfully "
> > Screen.Recording.2023-05-08.at.15.42.45.mov
> 
> changes requested /\ are simple to do now.
> 
> ideally I would like to see this flow: ( @luizstacio please check what you think before @tomiiide develops it) maybe we can do it in other PR as well:
> 
> 1A- user clicks "Add Account" 1B- user clicks "Import from Private" 1B- user informs private key and click ok 2- both 1A and 1B will be redirect to this page:
> 
> * show loading phrase
> * show account loading component
>   example:
>   image
> * Show a success screen
>   example:
>   image

I don't think we need this flow, as the end goal for the user it to add the account and go back to the home screen as fast as possible. Adding this screens would make the flow longer and slow.

luizstacio:
> @LuizAsFight - Wouldn't it make sense to ask user confirmation in a dialog before proceeding with creation of an account, similar to what we've for delete account currently? cc: @luizstacio

As after adding  the user can delete we don't need a confirmation.

luizstacio (changes requested):
- Move the loader state to only show the loading on the button and not on the entire screen. On the button loading, add the text `Creating...`

LuizAsFight:
> > > 1- I think we should keep loading state only for the button here... no needs to put list on loading, as we're going directly to home right after the action. 2- we should show a toast `"Account X" created successfully "
> > > Screen.Recording.2023-05-08.at.15.42.45.mov
> > 
> > 
> > changes requested /\ are simple to do now.
> > ideally I would like to see this flow: ( @luizstacio please check what you think before @tomiiide develops it) maybe we can do it in other PR as well:
> > 1A- user clicks "Add Account" 1B- user clicks "Import from Private" 1B- user informs private key and click ok 2- both 1A and 1B will be redirect to this page:
> > 
> > * show loading phrase
> > * show account loading component
> >   example:
> >   image
> > * Show a success screen
> >   example:
> >   image
> 
> I don't think we need this flow, as the end goal for the user it to add the account and go back to the home screen as fast as possible. Adding this screens would make the flow longer and slow.

ok but I think we're missing feedback. a success toast will help.

can you also add after creating a network @tomiiide ?

LuizAsFight:
> @LuizAsFight - Wouldn't it make sense to ask user confirmation in a dialog before proceeding with creation of an account, similar to what we've for delete account currently? cc: @luizstacio

delete needs confirmation because it's a critical action, but for creating it's not needed imo

packages/app/src/systems/Account/services/account.ts
tomiiide: Wow, thanks. I didn't know we had this function, thanks.

luizstacio (approved):

packages/app/src/systems/Account/machines/accountsMachine.tsx
```diff
@@ -260,6 +293,39 @@ export const accountsMachine = createMachine(
...
+          });
+
+          // set as active account
+          account = await AccountService.setCurrentAccount({
+            address: account.address.toString(),
+          });
+
+          return account;
+        },
+      }),
```
luizstacio, line 328: We had another machine for this before, to avoid the action to be blocked if the account list is loading.

luizstacio (changes requested):

packages/app/src/systems/Account/machines/accountsMachine.tsx
tomiiide: hmm, okay, I didn't think of that, I thought we had a AddAccount machine because we had a AddAccount page for it.

tomiiide: Going to move this back now.

packages/app/src/systems/Overlay/components/OverlayDialog/OverlayDialog.tsx
```diff
@@ -19,7 +19,6 @@ export function OverlayDialog() {
       <Dialog.Content css={styles.content}>
         {/* Accounts */}
         {overlay.is('accounts.list') && <Accounts />}
-        {overlay.is('accounts.add') && <AddAccount />}
```
LuizAsFight, line 22: should also remove leftovers from 'accounts.add'.
also overlay events to call `openAccountsAdd`

LuizAsFight (changes requested):

luizstacio (approved):

Issue #557, opened 2023-02-09
Add copy button to toast with error


LuizAsFight:
### Motivation

Sometimes the error is too big to be readable inside a toast. We should add a copy button

### Usage example

![image](https://user-images.githubusercontent.com/8636507/217952389-e86a0f25-bbf1-4cc7-8277-7e3c4ca67efd.png)

SakshiShah29:
I would like to contribute to this issue. I understand the relevant changes needed and will work on them.

PR #744, opened 2023-05-11
feat(toast): copy button when toast message is too large

tibi77:
I tested this in another screen and it looks good because I don t know how to trigger an error inside the app.

Also, can i do a before and after selector inside the fuels CSS? and thus remove the toast.css file?


tibi77:
@LuizAsFight can you do your add reviewers magic in here as well ?

tibi77:
I can not link issues but this is related to #557

LuizAsFight (changes requested):
I think this should be done in `fuel-ui` side and benefit also other projects that might need a copy button in toast for wathever reason.

this PR and other project would just use an option 

for example
`toast.error(TEXT, { showCopy: true })` 


this is tracked in [FRO-129](https://linear.app/fuel-network/issue/FRO-129/allow-toast-to-have-copy-button)
for now I will add `blocked` here until waiting for the fuel-ui implementation

luizstacio:
Closing this to move it to the fuel-ui repo.

PR #747, opened 2023-05-12
feat: add confirmation text for logout

### LINKED ISSUE #494: Logout confirmation
[FRO-66](https://linear.app/fuel-network/issue/FRO-66/logout-confirmation)

As logout is an action that can lead users to lose the seed phrase, we should add an input forcing the user to type before allowing the logout.

- Add confirmation input: "I have my recovery phrase.”

tibi77:
This should close #494, added the tests as well


packages/app/src/systems/Account/pages/Logout/Logout.tsx
```diff
@@ -42,12 +74,33 @@ export const Logout = () => {
...
       <Dialog.Footer>
         <Button
           aria-label="Logout"
           onPress={handlers.logout}
           isLoading={isLoading}
-          isDisabled={isLoading}
+          isDisabled={
+            isLoading ||
+            form.getValues().logoutConfirmation !== logOutConfirmationPhrase
+          }
```
tomiiide, line 103: can we move this into a const `isLogoutButtonDisabled` ?
```suggestion
          isDisabled={isLogoutButtonDisabled}
```

packages/app/src/systems/Account/pages/Logout/Logout.tsx
```diff
@@ -1,10 +1,42 @@
...
+  const form = useForm<FormValues>({
+    resolver: yupResolver(schema),
+    reValidateMode: 'onChange',
+    mode: 'onChange',
+    defaultValues: {
+      logoutConfirmation: '',
+    },
+  });
+  const { control } = form;
   return (
```
tomiiide, line 40:
```suggestion
 const isLogoutButtonDisabled = isLoading ||
            form.getValues().logoutConfirmation !== logOutConfirmationPhrase
  return (
```

tomiiide (changes requested):
Great work @tibi77.

Can you make the little code clean up. thanks.

tomiiide (dismissed):
looks good.

tibi77:
@tomiiide if you could merge, for some reason the vercel deploy does not work for me. I need to approve it from the org

sarahschwartz (changes requested):
Thanks @tibi77 ! Can you also remove the scroll for the dialog description and adjust the padding so everything fits on the screen?

tibi77:
for sure @sarahschwartz looking into it now.

luizstacio (changes requested):
I think we can do some changes to improve the UI;
- The confirmation could be done on a Confirmation PopUp
- If inside the screen, all the components needs to fix on a single scroll and the content should be inside the same padding, currently it's looking 

Screenshot 2023-05-22 at 3 01 38 PM

luizstacio (changes requested):
I think we can do some changes to improve the UI;
- The confirmation could be done on a Confirmation PopUp
- If inside the screen, all the components needs to fix on a single scroll and the content should be inside the same padding, currently it's looking 

Screenshot 2023-05-22 at 3 01 38 PM

packages/app/src/systems/Account/pages/Logout/Logout.tsx
```diff
@@ -42,12 +78,30 @@ export const Logout = () => {
...
+        render={({ field }) => (
+          <Input>
+            <Input.Field
+              {...field}
+              aria-label="Confirm Log Out"
+              placeholder={logOutConfirmationPhrase}
+            />
+          </Input>
+        )}
+      />
```
LuizAsFight, line 98: why is the controlled field outside of <Dialog.Description>
if you put this in the right place, layout will look ok already

packages/app/src/systems/Account/pages/Logout/Logout.tsx
```diff
@@ -42,12 +78,30 @@ export const Logout = () => {
...
         </Card>
       </Dialog.Description>
+      <ControlledField
+        control={control}
+        name="logoutConfirmation"
+        label={
+          <p>
+            Write <i>{logOutConfirmationPhrase}</i> to logout
+          </p>
+        }
```
LuizAsFight, line 88: we don't need to wrap it with `<p>.`

LuizAsFight, line 88:
```suggestion
        label={`Write "${logOutConfirmationPhrase}" to confirm`}
```

LuizAsFight (changes requested):

LuizAsFight:
@tibi77 the code changes I asked should be enough to fix @luizstacio 's changes requested

packages/app/src/systems/Account/pages/Logout/Logout.tsx
tibi77: thanks

packages/app/src/systems/Account/pages/Logout/Logout.tsx
tibi77: thanks; did the changes

tibi77:
image
thanks @sarahschwartz @LuizAsFight @luizstacio for the comments;
did the requested changes

sarahschwartz:
The prompt shows up twice now: 
wallet

tibi77:
@sarahschwartz weirdly i can not reproduce your behaviour; are you accessing the menu from a different menu or?
from the settings one ? 
image

tibi77:
for reference that's what i m seeing in the logout menu 
image

tibi77:
@sarahschwartz done with the nit :)

sarahschwartz (approved):
LGTM 💚

PR #749, opened 2023-05-14
fix: edit account refresh account list

eswarasai:
[FRO-111](https://linear.app/fuel-network/issue/FRO-111/after-editing-a-account-the-account-list-should-be-refetch)


tomiiide (approved):
LGTM!

PR #751, opened 2023-05-15
fix: don't hide action buttons when transaction window is zoomed in

### LINKED ISSUE #350: Failure to display transaction confirmation or rejection buttons

0.0.1

### What is happening?

I used wallet  version 0.0.1 in Chrome browser version 108.0.5359.98 (Official Build) (64-bit) and the bug that I saw was that the size of the extension page was not complete when i try  confirming the transaction

### How it should behave?

I used wallet  version 0.0.1 in Chrome browser version 108.0.5359.98 (Official Build) (64-bit) and the bug that I saw was that the size of the extension page was not complete when i try  confirming the transaction

Windows Desktop

### What browser?

Chrome

108.0.5359.98 (Official Build) (64-bit)

tomiiide:
Closes #350 / [FRO-80](https://linear.app/fuel-network/issue/FRO-80/failure-to-display-transaction-confirmation-or-rejection-buttons)

### DIFF
--- a/packages/app/popup.html
+++ b/packages/app/popup.html
@@ -11,7 +11,7 @@
       body {
         height: 600px;
         width: 350px;
-        overflow: hidden;
+        overflow: overlay;
         margin: 0;
         padding: 0;
         background-color: #101112;
--- a/.changeset/angry-mayflies-taste.md
+++ b/.changeset/angry-mayflies-taste.md
@@ -1,0 +1,5 @@
+---
+'fuels-wallet': minor
+---
+
+Fixed edit account not updating the account list
--- a/.changeset/cool-spoons-prove.md
+++ b/.changeset/cool-spoons-prove.md
@@ -1,0 +1,5 @@
+---
+'fuels-wallet': patch
+---
+
+Make the body overflow scroll for when the pop up window is zoomed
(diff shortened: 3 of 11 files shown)


packages/app/popup.html
```diff
@@ -11,7 +11,7 @@
       body {
         height: 600px;
         width: 350px;
-        overflow: hidden;
+        overflow: overlay;
```
LuizAsFight, line 14: pretty subtle change but can have a lot of impact. did you test this in windows ?

I was reading and:

> On the other hand, with overflow: overlay, a scrollable overlay is created regardless of whether the content fits within the defined dimensions of the element or not. This means that even if the content does not exceed the dimensions of the element, a scrollbar will still be created, and the content will be partially hidden under the overlay.
> So, in summary, overflow: auto only creates a scrollbar if necessary, while overflow: overlay always creates a scrollable overlay, regardless of whether the content fits within the defined dimensions or not.


Looks like will always show a scrollbar, but in mac:
- will not show always 
- shows a cool scrollbar

but in windows probably:
- will show always
- will show that ugly  default scrollbar, which we normally use `coreStyles.coreStyles()` to have a cool scrollbar always instead.

LuizAsFight (changes requested):

PR #753, opened 2023-05-18
feat: index all pages on docsearch

luizstacio:
Add all links for docsearch to index, the current approach was not indexing all pages as docsearch was not able to detect all paths related to `/`.


sarahschwartz:
Is there a reason the reference page is left out?

luizstacio:
> Is there a reason the reference page is left out?

Thanks for the review, I didn't notice that all links where different. I include other ones that were also missing.

sarahschwartz (approved):

PR #784, opened 2023-06-10
Reset auto-lock timer on user interaction

eswarasai:
[FRO-109](https://linear.app/fuel-network/issue/FRO-109/auto-lock-should-reset-the-timer-on-every-interaction-of-the-user)

### Description
- [x] Reset auto-lock timer on every user interaction, if wallet is unlocked
- [x] Moved `AUTO_LOCK_IN_MINUTES` to an environment variable


packages/app/src/config.ts
```diff
@@ -1,3 +1,5 @@
+import * as process from 'process';
```
luizstacio, line 1: Why we need this import process here?

packages/app/src/systems/CRX/background/services/VaultService.ts
```diff
@@ -48,7 +66,6 @@ export class VaultService extends VaultServer {
   async autoUnlock() {
     const secret = await loadSecret();
     if (secret) {
-      // Unlock vault directly without saving a new timestamp
```
luizstacio, line 51: We should keep this comment

packages/app/src/systems/CRX/background/services/VaultService.ts
```diff
@@ -32,6 +38,18 @@ export class VaultService extends VaultServer {
     this.emitLockEvent();
   }
 
+  async isLocked(): Promise<boolean> {
+    const isWalletLocked = await super.isLocked();
+    if (!isWalletLocked) {
+      const timer = await getTimer();
+      if (timer) {
+        // Saving a new timestamp for wallet auto lock
+        saveTimer(AUTO_LOCK_IN_MINUTES);
```
luizstacio, line 47: I think we could change the `saveSecret` method logic to save `lockTime` on the side of the `time` on the session. And then use this field value on a method `resetTimer`. This will reduce the need to pass the AUTO_LOCK_IN_MINUTES around and centralize just on `saveSecret`.

packages/app/.env.production
```diff
@@ -1,4 +1,5 @@
 VITE_FUEL_PROVIDER_URL=https://beta-3.fuel.network/graphql
 VITE_FUEL_FAUCET_URL=https://faucet-beta-3.fuel.network/
 VITE_MNEMONIC_WORDS=12
-VITE_FAUCET_RECAPTCHA_KEY=6Ld3cEwfAAAAAMd4QTs7aO85LyKGdgj0bFsdBfre
\ No newline at end of file
+VITE_FAUCET_RECAPTCHA_KEY=6Ld3cEwfAAAAAMd4QTs7aO85LyKGdgj0bFsdBfre
+VITE_AUTO_LOCK_IN_MINUTES=10
```
luizstacio, line 5: No need for empty line between envs.

luizstacio (changes requested):

Issue #389, opened 2022-12-21
Enable developers to request add new network


luizstacio:
[FRO-84](https://linear.app/fuel-network/issue/FRO-84/enable-developers-to-request-add-new-network)

Currently, a DApp cannot request the Wallet User to connect to a different network. We should add this option to the SDK. This work also includes the following;

- [ ] Create a new method `addNetwork` on the `SDK`.
- [ ] Create a screen for New Network Request.

PR #793, opened 2023-06-21
feat: create addNetwork request method on sdk

eswarasai:
[FRO-84](https://linear.app/fuel-network/issue/FRO-84/enable-developers-to-request-add-new-network)
Fixes: #389 

### Description
Implemented `addNetwork()` method on Fuel SDK enabling developers to add network

add-network-request

### TODO
- [x] Add storybook for `AddNetworkRequest` screen


packages/app/src/systems/DApp/pages/AddNetworkRequest/AddNetworkRequest.tsx
```diff
@@ -0,0 +1,75 @@
...
+          <Card.Header space="compact">
+            Review the Network to be added:
+          </Card.Header>
+          <Card.Body css={styles.cardContentSection}>
+            <Text as="h2" fontSize="lg" color="intentsBase12">
+              {network.name}
+            </Text>
+            <Text fontSize="sm">{network.url}</Text>
+          </Card.Body>
+        </Card>
```
LuizAsFight, line 35: this is looking different from the one in `Add Network` screen.

can we create a new component "ReviewNetworkCard" to make sure we use same styling and have more consistency ?

packages/docs/examples/AddNetwork.tsx
```diff
@@ -0,0 +1,102 @@
...
+import { useIsConnected } from '../src/hooks/useIsConnected';
+import { useLoading } from '../src/hooks/useLoading';
+
+export function AddNetwork() {
+  const [fuel, notDetected] = useFuel();
+  const [isConnected] = useIsConnected();
+  const [network, setNetwork] = useState<Network>({
+    name: '',
+    url: '',
+  });
```
LuizAsFight, line 19: can we use Localhost as default network ?

```suggestion
  const [network, setNetwork] = useState<Network>({
    name: 'Localhost',
    url: 'http://localhost:4000/graphql',
  });
```

LuizAsFight (changes requested):
when adding a new network from docs, it doesn't show when opening network list. I needed to open network list 2 times for it to show.

then I removed the network from the list, and tried to add again through doc.

Adding worked, but now the new network doesn't show at all in the list, and the requests are going to Localhost network I added, although wallet is showing current network as Beta-3.

should investigate how this is happening and totally remove the possibility of consume one network, but show another one as current network.
Also would be nice to have e2e test covering this scenario of :
1- add network from dapp
2- new network should show in network list first time it's opened
3- remove network in wallet list
4- add network from dapp again
5- new network should show in network list first time it's opened
6- requests should match current network URL

image
image

packages/docs/docs/dev/networks.mdx
```diff
@@ -13,6 +13,14 @@ To detect the current network the user's wallet is connected to, use the `networ
 
 <Examples.Network />
 
+## Adding custom Network
```
LuizAsFight, line 16: should also add new methods to `reference.mdx` and `Fuel.test.ts`

LuizAsFight: also add new method `networks` here

LuizAsFight (changes requested):
nice work @eswarasai , few changes but overall looking pretty solid

packages/app/src/systems/DApp/hooks/useAddNetworkRequest.tsx
```diff
@@ -0,0 +1,52 @@
...
+  /* store.useUpdateMachineConfig(Services.txRequest, {
+    actions: {
+      notifyRefreshNetworks () {
+        store.refreshNetworks();
+      },
+      notifyUpdateAccounts () {
+        store.updateAccounts();
+      },
+    },
+  }); */
```
luizstacio, line 25: You can remove this code?

packages/app/src/systems/Network/hooks/useNetworks.ts
```diff
@@ -31,6 +32,10 @@ export function useNetworks() {
     selectors.selectedNetwork
   );
 
+  useEffect(() => {
+    store.refreshNetworks();
+  }, []);
```
luizstacio, line 37: Why we need this refresh here?

luizstacio (changes requested):
- When adding a URL that already exists the SDK should throw an error before opening the UI for the user. This validation needs to be on the same level of url format.

packages/app/src/systems/Network/hooks/useNetworks.ts
eswarasai: Just to ensure that the `networks` in the `store` are being refreshed and up-to-date with the data stored in the `IndexedDB`
cc: @LuizAsFight

luizstacio: @LuizAsFight do we have some cache on the network list? If not I don't see why it would be needed.

LuizAsFight: @luizstacio 
this is just to make sure that when network list opens, network list will be correct

as network machine is global, if some other machine forget to call `notifyRefreshNetworks` after change networks, the list will be incorrect.

packages/app/src/systems/CRX/background/services/BackgroundService.ts
```diff
@@ -341,4 +343,39 @@ export class BackgroundService {
...
+    if (!isValidNetworkUrl(network.url)) {
+      throw new Error('Invalid network URL');
+    }
+
+    const networkExists = await NetworkService.getNetworkFromUrl({
+      url: network.url,
+    });
+    if (networkExists) {
+      throw new Error('Network already exists');
+    }
```
LuizAsFight, line 361: this network validation are getting too spread, we should create a function in NetworkService

`NetworkService.assertAddNetwork`

this method will have the necessary validations to add a network.

it should be reused in other places to clean

packages/app/src/systems/Network/services/network.ts
```diff
@@ -51,11 +54,12 @@ export class NetworkService {
   static addNetwork(input: NetworkInputs['addNetwork']) {
```
LuizAsFight, line 54: this method should use new created `NetworkService.assertAddNetwork`

packages/app/src/systems/DApp/machines/addNetworkRequestMachine.tsx
```diff
@@ -0,0 +1,148 @@
...
+            });
+            return selectedNetwork;
+            // eslint-disable-next-line @typescript-eslint/no-explicit-any
+          } catch (error: any) {
+            if (error?.message.includes('uniqueness')) {
+              throw new Error('This network Name or URL already exists');
+            }
+            throw error;
+          }
+        },
```
LuizAsFight, line 138: clean it as now `NetworkService.addNetwork` will do the validations using `NetworkService.assertAddNetwork`

LuizAsFight (changes requested):

packages/docs/docs/dev/reference.mdx
```diff
@@ -45,6 +45,13 @@ Return the current account being used in the wallet application.
   testCase="currentAccount"
 />
 
+### Add Network
```
LuizAsFight, line 48: should also add List Networks, and Current Network

LuizAsFight (changes requested):

packages/app/src/systems/Network/hooks/useNetworks.ts
luizstacio: I don't think we should cover forgetting `notifyRefreshNetworks`. This needs to be called for an application to run correctly. And as we don't have caches every time the application is open, it should refresh.

LuizAsFight: I don't see what's the problem, a refetch when opening a list is a pretty common behavior

LuizAsFight (approved):
lgtm 🏖️

packages/docs/docs/dev/networks.mdx
```diff
@@ -13,6 +13,14 @@ To detect the current network the user's wallet is connected to, use the `networ
...
+
+To add custom Network, use the `wallet.addNetwork` method and pass in the `Network` you want to add.
+
+<CodeImport file="../../examples/AddNetwork.tsx" />
```
luizstacio, line 20: I think it is missing comment tag `network `

luizstacio (changes requested):

luizstacio (approved):

LuizAsFight (approved):

PR #801, opened 2023-07-07
fix: connect multiple accounts even if current account is connected

eswarasai:
[FRO-242](https://linear.app/fuel-network/issue/FRO-242/when-click-to-connect-show-popup-to-connect-accounts)


luizstacio (changes requested):

packages/app/src/systems/DApp/machines/connectRequestMachine.ts
```diff
@@ -183,26 +193,39 @@ export const connectRequestMachine = createMachine(
...
             },
           });
         },
       }),
     },
     guards: {
-      connectionConnected: (_, ev) => {
-        return !!((ev.data?.accounts.length || 0) >= 1);
+      connectionConnected: (ctx, ev) => {
+        return (ev.data?.accounts.length || 0) === ctx.totalAccounts;
```
matt-user, line 228: @luizstacio should we also add a check here that `ctrx.totalAccounts >= 1`?

packages/app/src/systems/DApp/machines/connectRequestMachine.ts
```diff
@@ -183,26 +193,39 @@ export const connectRequestMachine = createMachine(
...
+            });
           }
           return ConnectionService.addConnection({
             data: {
               origin: input.origin,
-              accounts: input.accounts,
-              favIconUrl: input.favIconUrl,
-              title: input.title,
+              accounts: input.selectedAddresses,
+              favIconUrl: input.favIconUrl!,
```
matt-user, line 219: @luizstacio we should find some way to remove the `!` right?  The easiest way is to add a check such as
```
if (!input.favIconUrl || !input.title) throw Error ("...");
```
is there a better way

luizstacio: In some cases, it's ok to use it, but I think in this case is not required, it should accept undefined...

packages/app/src/systems/DApp/machines/connectRequestMachine.ts
luizstacio: yeah 
```suggestion
        return (ev.data?.accounts.length || 0) === ctx.totalAccounts && ctx.totalAccounts >= 1
```

luizstacio (approved):
LGTM

PR #803, opened 2023-07-07
feat: remove assets title from list

SilentCicero:
## Abstract

We should aim to reduce wasted valuable UI real estate as much as possible. Having a title that states "Assets" during the entire wallet lifecycle is not necessary and is self explanatory for the user.

Here i'm removing it, but overall, we should aim to remove things like this that are either "self explanatory" or "things I only need to see once".

Open to thoughts from the team.


luizstacio:
For me, it seems fine to remove it, or just make it smaller. cc: @kallemoen

PR #821, opened 2023-07-29
feat: update fuel-ui and other libs

LuizAsFight:


packages/app/src/systems/Asset/hooks/useAssetForm.tsx
```diff
@@ -11,7 +11,7 @@ import type { Maybe } from '~/systems/Core';
 export type AssetFormValues = {
   name: string;
   assetId: string;
-  imageUrl: string;
+  imageUrl: string | undefined;
```
luizstacio, line 14:
```suggestion
  imageUrl?: string;
```

packages/app/src/systems/Unlock/hooks/useUnlockForm.tsx
```diff
@@ -12,7 +12,7 @@ const schema = yup
 export type UseUnlockFormReturn = ReturnType<typeof useUnlockForm>;
 
 export type UnlockFormValues = {
-  password: string;
+  password: string | undefined;
```
luizstacio, line 15:
```suggestion
  password?: string;
```

luizstacio: Why does this need to accept undefined?

luizstacio (changes requested):
It's showing some warning on the code also.

packages/app/src/systems/Asset/hooks/useAssetForm.tsx
LuizAsFight: on this way it breaks yupResolver typing
image

packages/app/src/systems/Unlock/hooks/useUnlockForm.tsx
LuizAsFight: because now for props inside of schema object that are not `required`,  `yupResolver` will accept only together with `undefined` typing

luizstacio (approved):

PR #846, opened 2023-08-29
chore: update dropdown color

luizstacio:


LuizAsFight (approved):
this works for now but I think we should fix it in fuel-ui to avoid this problem everywhere else

Issue #506, opened 2023-01-24
[RFC] Fuel Wallet Connectors


luizstacio:
# Fuel Wallet Connector

## Objective
Enable connectors that are not extensions to seamlessly integrate with Fuel Wallet's multiple connector standard, thus enhancing application-DApp interoperability.

## Background

As blockchain Wallet evolved, each wallet chose its protocol, combining different integration methods. To solve these problems, we saw many abstractions tooling to make a single interface to integrate with multiple Wallet types.

On Fuel, we start with a Connectors protocol to enable multiple Wallet Providers to integrate directly on the Fuel Wallet SDK,  removing the need for other abstractions when developing Applications.

The current implementation of the Fuel Wallet Connectors only works with browser extensions. This proposal aims to change that, opening up more customization possibilities. 

## Problem Statement

_How can we improve the methods through which applications interact with different types of connectors, including those that are not browser extensions?_

## Definitions

- Fuel Wallet SDK: Manager component used for developers to integrate with the Fuel Wallet Connectors.
- Fuel Wallet Connector: Component provided by a Wallet Application that implements the logic for the communication between the Fuel Wallet SDK and the Wallet Application.

## Goals

- Implement a set of methods and events that will standardize interactions between applications and connectors.
- Enable application information and installation instructions through connector metadata.
- Allow Fuel Wallet SDK to detect installed Wallets automatically.
- Remove abstractions for the communication between SDK and Fuel Wallet Connectors.

#### Methods
- The system will expose methods like ping, version, isConnected, etc. to enable standardized interaction.

#### Events
- Events like `accounts`, `currentAccount`, `connection`, etc., will be emitted to communicate changes.

#### Connector Metadata
- A metadata field for connectors will store name, image, and installation instructions.

#### Fuel Wallet SDK
- The SDK connects the connectors, to invoke the exposed methods and listen to various events.

#### Method list

| Method | Description | Params | Return |
| --- | --- | --- | --- |
| ping | It should return true if the connector is loaded in less than one second. |  | boolean |
| version | The current version and the network that this connector supports. |  | Application and network version |
| isConnected | The status of the current connection from the origin to the connector. |  | connection status |
| accounts | Accounts available to the authorized to the current connection. |  | Account address list |
| connect | Start the authorization flow for the current connection. |  | connection status |
| disconnect | Removes the authorization of the current connection. |  | connection status |
| signMessage | Start signMessage flow for the current connection. | address and message | message signature |
[...]

PR #861, opened 2023-09-06
fix: sdk connectors syncing

luizstacio:
Currently, when using the SDK directly instead of window. Fuel, the connectors are not updated once they are available. This fix creates a syncing handler to ensure data is bonded between both. This will probably improve the communication protocol refactor. https://github.com/FuelLabs/fuels-wallet/issues/506

### DIFF
--- a/packages/sdk/src/connections/WindowConnection.ts
+++ b/packages/sdk/src/connections/WindowConnection.ts
@@ -129,6 +129,28 @@ export class WindowConnection extends BaseConnection {
     window.postMessage(message, origin || window.origin);
   }
 
+  bindFuelConnectors(fuel: Window['fuel']) {
+    // Prevent binding to self if this happen the
+    // object would enter on a infinite loop
+    const isSelf = (fuel as WindowConnection) === this;
+    if (!fuel || isSelf) return;
+    // Bind to fuel events to
+    // sync with current instace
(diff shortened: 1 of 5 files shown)


packages/sdk/src/connections/WindowConnection.ts
```diff
@@ -129,6 +129,29 @@ export class WindowConnection extends BaseConnection {
...
+    // Prevent binding to self if this happen the
+    // object would enter on a infinite loop
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const isSelf = (fuel as any) === this;
+    if (!fuel || isSelf) return;
+    // Bind to fuel events to
+    // sync with current instace
+    fuel.on('connectors', (connectors) => {
+      this.connectors = connectors;
+      this.emit('connectors', connectors);
```
matt-user, line 142: what is the point of emitting the event here?  Isn't the event already emitted by the fact that we are handling it here?

pedronauck (approved):

packages/sdk/src/connections/WindowConnection.ts
luizstacio: The emit happens on the current instance. But the listener listens to the events on the `window.fuel` instance.

PR #867, opened 2023-09-15
feat: add `@fuel-wallet/react`

sdankel:
Closes https://github.com/FuelLabs/fuels-npm-packs/issues/20


packages/react/src/components/FuelProvider.tsx
```diff
@@ -0,0 +1,90 @@
+import type { Fuel } from '@fuel-wallet/sdk';
```
luizstacio, line 1: Can you apply the changes I made here? https://github.com/FuelLabs/fuels-portal/blob/a2c20159bd6b040d53e62d43f02885b38ccb8556/packages/sdk-react/src/components/FuelProvider.tsx

Only the change fuelQueryClient.fetchQuery to fuelQueryClient.invalidateQueries?

This will solve a problem where queries were triggered without generating errors. In case we are not using all the hooks.

packages/react/tsup.config.ts
```diff
@@ -0,0 +1,8 @@
+export default {
+  sourcemap: true,
```
luizstacio, line 2: Can you add here dts: true and remove `--dts` on the "build" command

packages/react/package.json
```diff
@@ -0,0 +1,42 @@
...
+      "dist"
+    ]
+  },
+  "scripts": {
+    "build": "pnpm ts:check && tsup --dts",
+    "ts:check": "pnpm tsc --noEmit"
+  },
+  "peerDependencies": {
+    "fuels": ">=0.57.0",
+    "@fuel-wallet/sdk": "workspace:*"
```
luizstacio, line 29: We can move `@fuel-wallet/sdk` into the `dependencies` section.

luizstacio (changes requested):

packages/react/tsup.config.ts
sdankel: Should I do this for `packages/sdk` as well?

LuizAsFight (approved):

luizstacio (approved):

PR #876, opened 2023-09-21
feat: update ts-sdk and move provider from sync to async

luizstacio:


matt-user (dismissed):
why was the __mock__/dapp-provider removed?

luizstacio:
> why was the **mock**/dapp-provider removed?

Was not in use

packages/app/playwright/commons/gas.ts
```diff
@@ -4,7 +4,7 @@ import type { Provider, BN } from 'fuels';
 // TODO: remove this function when SDK make transactions with correct gas configs
 export const getGasConfig = async (provider: Provider) => {
   const chain = await provider.getChain();
-  const nodeInfo = await provider.getNodeInfo();
+  const nodeInfo = await provider.fetchNode();
```
LuizAsFight, line 7: why you need to `fetchNode`? now when you do `await Provider.create()` it will do this request and cache for you. you can syncly access it:

```suggestion
  const nodeInfo = provider.getNode();
```

packages/app/playwright/crx/crx.test.ts
```diff
@@ -405,7 +405,7 @@ test.describe('FuelWallet Extension', () => {
 
             // TODO: remove this gas config once SDK fixes and start with correct values
             const chain = await wallet.provider.getChain();
-            const nodeInfo = await wallet.provider.getNodeInfo();
+            const nodeInfo = await wallet.provider.fetchNode();
```
LuizAsFight, line 408:
```suggestion
            const nodeInfo = wallet.provider.getNode();
```

packages/app/src/systems/Network/services/network.ts
```diff
@@ -138,14 +138,14 @@ export class NetworkService {
...
+    const provider = await Provider.create(input.providerUrl);
     return provider.getChain();
   }
 
-  static getNodeInfo(input: NetworkInputs['getNodeInfo']) {
-    const provider = new Provider(input.providerUrl);
-    return provider.getNodeInfo();
+  static async getNodeInfo(input: NetworkInputs['getNodeInfo']) {
+    const provider = await Provider.create(input.providerUrl);
+    return provider.fetchNode();
```
LuizAsFight, line 148:
```suggestion
    return provider.getNode();
```

packages/app/src/systems/Transaction/services/transaction.tsx
```diff
@@ -251,7 +253,7 @@ export class TxService {
   }
 
   static async fundTransaction(input: TxInputs['fundTransaction']) {
-    const { minGasPrice } = await input.wallet.provider.getNodeInfo();
+    const { minGasPrice } = await input.wallet.provider.fetchNode();
```
LuizAsFight, line 256:
```suggestion
    const { minGasPrice } = await input.wallet.provider.getNode();
```

packages/react/src/hooks/useNodeInfo.tsx
```diff
@@ -14,7 +14,7 @@ export const useNodeInfo = ({ version = '0.0.0' }: NodeInfoParams = {}) => {
   const { data: nodeInfo, ...query } = useQuery(
     [QUERY_KEYS.nodeInfo, provider?.url],
     () => {
-      return provider?.getNodeInfo();
+      return provider?.fetchNode();
```
LuizAsFight, line 17:
```suggestion
      return provider?.getNode();
```

LuizAsFight (changes requested):
replace also other occurrences of `fetchNode` ?

luizstacio:
> replace also other occurrences of `fetchNode` ?

We need to do it anyway, as the data is not auto-populated to all the places. But as we do at the beginning it will get from cache and not fetch the provider again.

LuizAsFight (approved):

PR #886, opened 2023-09-29
test: e2e contract test mint

matt-user:
closes FRO-596


packages/e2e-contract-tests/.eslintrc.cjs
```diff
@@ -0,0 +1,18 @@
+module.exports = {
```
luizstacio, line 1: I think we don't need another lint config on the project as it's going to use the root one.

packages/e2e-contract-tests/.gitignore
```diff
@@ -0,0 +1,27 @@
+# Logs
```
luizstacio, line 1: Also we have a root ignore file on this one we should add only the project specific files

packages/e2e-contract-tests/package.json
```diff
@@ -0,0 +1,37 @@
...
+  "scripts": {
+    "dev:e2e-contracts": "vite",
+    "build:web": "vite build",
+    "build:contracts": "pnpm exec fuels typegen -i ./contracts/custom_asset/out/debug/*-abi.json -o ./src/contracts",
+    "lint": "eslint . --ext ts,tsx --report-unused-disable-directives --max-warnings 0",
+    "preview": "vite preview"
+  },
+  "dependencies": {
+    "@fuel-ui/css": "0.21.0",
+    "@fuel-ui/react": "0.21.0",
```
luizstacio, line 14: To keep it at minimum deps we could use just react with default HTML styles

packages/e2e-contract-tests/playwright/e2e/MintAsset.test.ts
```diff
@@ -0,0 +1,59 @@
...
+    const walletPage = await getWalletPage(context);
+    // short address function copied from app package
+    await hasText(walletPage, shortAddress(assetId));
+
+    // test mint amount is correct
+    await hasText(walletPage, mintAmount);
+
+    // test gas fee is shown and correct
+    await hasText(walletPage, 'Fee (network)');
+    // 0.000000001 since we hardcode a gas price of 1 in our mint function
```
luizstacio, line 56: I didn't understand this part as the gasFee should be set on the chain level.

packages/e2e-contract-tests/src/vite-env.d.ts
```diff
@@ -0,0 +1,21 @@
...
+declare module '*?script&module' {
+  const src: string;
+  export default src;
+}
+
+// Add script  importing (CommonJS)
+declare module '*?script' {
+  const src: string;
+  export default src;
+}
```
luizstacio, line 21: Do we need all this envs definitions here?

packages/test-utils/package.json
```diff
@@ -0,0 +1,32 @@
+{
+  "name": "@fuel-wallet/test-utils",
```
luizstacio, line 2: We can add this as private for now, and release once we achieve a stable version

packages/test-utils/src/test-utils/fuelWallet.ts
```diff
@@ -0,0 +1,88 @@
+import { type BrowserContext, type Page } from '@playwright/test';
```
luizstacio, line 1: Nice work here 🤘

luizstacio (changes requested):

packages/e2e-contract-tests/playwright/e2e/MintAsset.test.ts
matt-user: yes gasFee is set on chain level, but we also use a gasPrice of 1 in the txParams of the mint function.  If the gasPrice is set higher then the network fee will also be higher right?

luizstacio (approved):

PR #887, opened 2023-10-01
fix: Remove unlisted dependency on React

dmihal:
Although `@fuel-wallet/types` does not specify React as a dependency, it still depends on react (specifically `@types/react`). If the library is used in a project that doesn't have `@types/react` installed, Typescript will give the following error:

```
node_modules/@fuel-wallet/types/src/error.ts:2:24 - error TS2307: Cannot find module 'react' or its corresponding type declarations.

2 import type React from 'react';
```

Given that the import is just a simple interface, I've copied the interface into the file and removed the import.

### DIFF
--- a/.changeset/nine-eggs-build.md
+++ b/.changeset/nine-eggs-build.md
@@ -1,0 +1,5 @@
+---
+'@fuel-wallet/types': patch
+---
+
+fix: remove unlisted dependency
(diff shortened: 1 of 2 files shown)


luizstacio:
@dmihal can you just add the changelog to the PR? `pnpm changeset add`

luizstacio (approved):

Issue #868, opened 2023-09-16
Add custom decimal sizes for custom assets


luizstacio:
### Motivation

When working with custom Assets, having different decimal sizes is typical. Right now, the wallet only supports `9` decimals.

### Usage example

A custom token with 3 decimals, or a custom token with a 10 decimals.

### Possible implementations

To do;
- Add a decimal field on the Asset on the Create Custom Asset form.
- Add the new decimal format on the inputs and places where we display amounts. If a component does not exist to abstract this, we should create it to avoid formating it at all places.
- Add tests to cover this on Transfer and TX History.

PR #894, opened 2023-10-04
feat: add decimal size for custom assets

luizstacio:
close: #868


packages/app/src/systems/Core/utils/math.ts
```diff
@@ -2,8 +2,9 @@ import type { BNInput } from 'fuels';
 import { bn } from 'fuels';
 import { MAX_FRACTION_DIGITS } from '~/config';
 
-export function formatAmount(amount?: BNInput) {
+export function formatAmount(amount?: BNInput, units?: number) {
   return bn(amount).format({
     precision: MAX_FRACTION_DIGITS,
+    units,
```
LuizAsFight, line 8: we don't need this function, can use directly `format` or `formatUnits` from `fuels`

LuizAsFight:
- card looking weird
- we could include decimal information here

image

LuizAsFight:
the amount that I put in doc, doesn't reflect correctly in wallet. I guess need to parse it before sending. (user should know the asset decimal beforehand?

image

luizstacio:
> the amount that I put in doc, doesn't reflect correctly in wallet. I guess need to parse it before sending. (user should know the asset decimal beforehand?
> 
> image

Yeah good catch we probably need to query the assets info on the docs to make the input better. But overall the application also needs to know this info to show the input with the correct format.

packages/app/src/systems/Core/utils/math.ts
luizstacio: Yeah, we should do some refactoring on the code base in the near future.

LuizAsFight (approved):

PR #896, opened 2023-10-05
feat: remove fuel window

luizstacio:
In this PR, we introduce some break changes as we remove the `window.fuel`. The idea of removing the instance from the window is to incentivize users to use directly the `@fuel-wallet/sdk` on DApps.

The main break changes are;
- Removal of the `window.fuel` instance.
- `createConnector` method is used for wallets to connect with the current instance. Even if the API signature didn't change,   the implementation is different and relies on window events for communicating that the connector is available.
- Move events from `document` to the `window`.
- Remove `useFuelWindow` from the `@fuel-wallet/react`.

To do;
- Update docs removing the use of the `window.fuel`.


luizstacio:
Close in favor of: #506
By finalizing the new communication protocol, `window.fuel` will also be removed, and we will avoid having two break changes.

PR #908, opened 2023-10-11
test: different asset configs

matt-user:
closes FRO-628


packages/e2e-contract-tests/playwright/e2e/MintAsset.test.ts
```diff
@@ -42,4 +44,68 @@ test.describe('Mint Assets', () => {
...
+    await hasText(walletPage, 'Fee (network)');
+    // 0.000000001 since we hardcode a gas price of 1 in our mint function
+    await hasText(walletPage, '0.000000001 ETH');
+
+    await walletApprove(context);
+
+    const name = 'Token';
+    const symbol = 'TKN';
+    // Add the unknown asset to the wallet
+    await addAsset(context, assetId, name, symbol, Number(decimals));
```
LuizAsFight, line 89: can we do the `addAsset` before sending the contract transaction? on this way we can validate both screens with same asset configuration

LuizAsFight (changes requested):

packages/e2e-contract-tests/playwright/e2e/MintAsset.test.ts
LuizAsFight: what's the point of validating the first screen with unknown asset and incorrect decimal config? and then after approve validate with the asset added ?

matt-user: my thinking for validating the unknown screen is ensuring that the wallet doesn't show any asset data before the asset is added

LuizAsFight: the first test already validates the scenario for unknown asset... this new test should validating the scenario for known asset

package.json
```diff
@@ -59,7 +59,7 @@
     "nodes:reset": "run-s nodes:clean nodes:up",
     "packages:version": "node ./scripts/version.js",
     "prettier:check": "prettier --check . !./packages/docs",
-    "prettier:format": "prettier --write .",
+    "prettier:format": "prettier --write . && pnpm -r --filter=docs exec prettier --write .",
```
luizstacio, line 62: why we have to change this configuration here?

luizstacio (changes requested):

package.json
matt-user: otherwise prettier uses a later version on the docs that causes the build to fail

packages/test-utils/src/test-utils/fuelWallet.ts
```diff
@@ -88,6 +96,94 @@ export async function walletApprove(context: BrowserContext) {
   await approveButton.click();
 }
 
+export async function addAssetById(
```
LuizAsFight, line 99:
```suggestion
export async function addAssetThroughSettings(
```

packages/test-utils/src/test-utils/fuelWallet.ts
```diff
@@ -88,6 +96,94 @@ export async function walletApprove(context: BrowserContext) {
...
+  const assetDecimalsInput = walletPage.getByLabel('Asset decimals');
+  await assetDecimalsInput.fill(decimals.toString());
+  const assetImageUrlInput = walletPage.getByLabel('Asset image Url');
+  await assetImageUrlInput.fill(imageUrl || '');
+
+  const saveButton = getButtonByText(walletPage, 'Save');
+  await saveButton.click();
+}
+
+export async function addAsset(
```
LuizAsFight, line 144:
```suggestion
export async function addAssetFromHomeBalance(
```

LuizAsFight (changes requested):
also all changes related to `packages/docs` looks to be orthogonal to this PR purpose. if yes, we should revert it and do in a separate PR if it's really needed

matt-user:
> also all changes related to `packages/docs` looks to be orthogonal to this PR purpose. if yes, we should revert it and do in a separate PR if it's really needed

yes they are mostly orthogonal.  They were needed originally, because my code was failing prettier:check and I needed to use prettier:format which broke the build without these orthogonal changes.  Now that my code is formatted correctly We could add these changes in a separate pr.

LuizAsFight (approved):

luizstacio (approved):

PR #936, opened 2023-10-19
fix: able connect to receive void args

### LINKED ISSUE #892: useConnect type error non-required param

0.13.0

### What is happening?

`const { connect } = useConnect();
async function handleClick() {
    await connect();
}`
Error:
TS2554: Expected 1-2 arguments, but got 0.

The connect function does not have required but optional arguments, however the type enforces it.

### steps to reproduce

```
const { connect } = useConnect();
async function handleClick() {
    await connect();
}
```
:warning: Then, I get this TS error
TS2554: Expected 1-2 arguments, but got 0.
It seems that no argument is required
```
// packages/react/src/hooks/useConnect.tsx
// ...
    mutationFn: async (connectorName?: string) => {
   if (connectorName) {
// ...
```
type coercion, this works:
`await connect(""); // empty string is a falsy value`

### How it should behave?

await connect(); // no type error

None

### What browser?

None

_No response_

Another-DevX:
close #892

### DIFF
--- a/packages/react/src/hooks/useConnect.tsx
+++ b/packages/react/src/hooks/useConnect.tsx
@@ -7,7 +7,7 @@ export const useConnect = () => {
   const { fuel } = useFuel();
 
   const { mutateAsync, ...mutateProps } = useMutation({
-    mutationFn: async (connectorName?: string) => {
+    mutationFn: async (connectorName: string | void) => {
       if (connectorName) {
         localStorage.setItem(CONNECTOR_KEY, connectorName);
         await fuel?.selectConnector(connectorName);


Another-DevX:
> close #892

The original mutationFn function was typed to expect a string argument, leading to a TypeScript type error when called without an argument. TypeScript expects that a function with a defined parameter will receive an argument during invocation; violating this expectation results in a type error. One might consider using connectorName?: string to mark the argument as optional, but this approach is incompatible with the type signature expected by React Query's useMutation, which leads to additional type errors. Therefore, changing the parameter type to string | void is the most effective way to both allow the function to be called without an argument and to maintain compatibility with React Query, resolving the initial type error and ensuring type safety.

luizstacio:
This is fixed on the new version of the sdk. Check branch `sdk-v2`

PR #939, opened 2023-10-20
feat: upgrade fuel-core

LuizAsFight:
Closes FRO-674

this PR:

- update wallet to fuel-core 0.20.7
- change chainConfig to be equal beta-4
- fix e2e contract tests


packages/app/src/systems/SignUp/components/CreatePassword/CreatePassword.tsx
```diff
@@ -54,7 +54,7 @@ export function CreatePassword({
   const form = useForm<CreatePasswordValues>({
     resolver: yupResolver(schema),
     reValidateMode: 'onChange',
-    mode: 'onChange',
+    mode: 'onBlur',
```
matt-user, line 57: How does this fix the password input showing incorrect error feedback?  All this changes is when the form values are validated right?

LuizAsFight: when you type a correct password, it shows incorrect (red border) in the field. although the strength indicator shows strong password, and the button works. Was just a rendering problem which was fixed changing this to `onBlur`

matt-user (approved):

PR #940, opened 2023-10-20
fix: showing assets with no amount

### LINKED ISSUE #931: [Withdraw Token] Approve tx showing empty operation
### Motivation

In the first transaction of withdrawing erc20 token, this is the approval screen

the second operation sending 0.000 ETH should not be shown

image

### Usage example

_No response_

### Possible implementations

_No response_

matt-user:
closes FRO-592
closes #931


LuizAsFight (changes requested):
I think this issue should be fixed on sdk side, txSummary should not return asset with value zero inside of operations,

you wanna work on a fix there ?

packages/app/src/systems/Asset/components/AssetsAmount/AssetsAmount.tsx
```diff
@@ -25,38 +25,50 @@ export const AssetsAmount: AssetsAmountComponent = ({
...
-                color: '$intentsError10',
-                fontSize: '$sm',
-                fontWeight: '$normal',
-              }}
-            >
-              (not enough balance)
-            </Text>
+          {!allEmptyAmounts && (
+            <Box.Stack gap="$2">
+              {amounts.map(
```
LuizAsFight, line 58: this can still show asset with zero value

if has 2 assets and only 1 of them is zero, wallet will show the 2 assets

LuizAsFight (changes requested):
even fixing on sdk side, we can go on with this PR , but need to fix the comment made

packages/app/src/systems/Asset/components/AssetsAmount/AssetsAmount.tsx
matt-user: It won't show the 1 zero asset bc of line 60

matt-user:
> I think this issue should be fixed on sdk side, txSummary should not return asset with value zero inside of operations,
> 
> you wanna work on a fix there ?

I don't think this is a bug in the ts-sdk.  The operation is a valid message output operation with an asset amount of 0, but I'll ask the sdk team

LuizAsFight (changes requested):
should hide the zeroed asset, not the whole operation
image

LuizAsFight (approved):

PR #951, opened 2023-10-25
fix: isConnected returning true for not connected accounts

### LINKED ISSUE #929: isConnected returning true even if currentAccount is not connected
When completing this task, fix it here according to the comment: 
useBridgeTxs line 43. // TODO

matt-user:
closes #929

### DIFF
--- a/packages/app/src/systems/CRX/background/services/BackgroundService.ts
+++ b/packages/app/src/systems/CRX/background/services/BackgroundService.ts
@@ -181,7 +181,13 @@ export class BackgroundService {
   }
 
   async isConnected(_: unknown, serverParams: EventOrigin) {
-    return !!serverParams.connection;
+    const currentAccount = await AccountService.getCurrentAccount();
+    if (!serverParams.connection) {
+      return false;
+    }
+    return serverParams.connection.accounts.includes(
+      currentAccount?.address || ''
+    );
   }
 
   async connect(_: JSONRPCParams, serverParams: EventOrigin) {
--- a/.changeset/hot-pants-relax.md
+++ b/.changeset/hot-pants-relax.md
@@ -1,0 +1,5 @@
+---
+'fuels-wallet': patch
+---
+
+fix: isConnected returning true for non-connected accounts


packages/app/src/systems/CRX/background/services/BackgroundService.ts
```diff
@@ -181,7 +181,13 @@ export class BackgroundService {
...
 
   async isConnected(_: unknown, serverParams: EventOrigin) {
-    return !!serverParams.connection;
+    const currentAccount = await AccountService.getCurrentAccount();
+    if (!serverParams.connection) {
+      return false;
+    }
+    return serverParams.connection.accounts.includes(
+      currentAccount?.address || ''
+    );
```
luizstacio, line 190: We should change this implementation to add an optional param to isConnected.

The idea of the isConnected method is not connected to the current account but to the state of the connection.

Another way to know if an account is connected or not is by checking on the accounts list.

luizstacio (changes requested):

packages/app/src/systems/CRX/background/services/BackgroundService.ts
luizstacio: The current behavior is not a bug but an expected behavior. cc: @LuizAsFight

LuizAsFight: what's the benefit of knowing that the account is connected but you can't do nothing as the current account is not connected? on this `isConnected` has no use as you can't trust it to show "connect" button for example.

luizstacio: You can request a signature or send transactions to any connected account.

LuizAsFight: @matt-user we can close this PR then

matt-user: @LuizAsFight ok I'll close this PR and refactor the code on the bridge

matt-user:
closed as this is not a bug, but a feature

PR #960, opened 2023-10-30
refactor: test utils

### LINKED ISSUE #957: Refactor and improve wallet test-utils
We need to refactor the test-utils package in the wallet so it can be used in other frontend projects.

Improvements:

- refactor the package to make it class based instead of a file of functions?
- add account
- add network
- connect multiple accounts
- make the functions more robust (ex: work even when the wallet page is not at the default page)

matt-user:
closes #957


packages/docs/docs/dev/test-utils.mdx
```diff
@@ -0,0 +1,134 @@
+---
+title: Wallet Test Utils
```
luizstacio, line 2: I think would make sense to change for `Playwright Utils`

packages/test-utils/package.json
```diff
@@ -1,6 +1,5 @@
 {
   "name": "@fuel-wallet/test-utils",
```
luizstacio, line 2: I think we should make the name `@fuel-wallet/playwright-utils`?

luizstacio (changes requested):

packages/e2e-contract-tests/playwright/e2e/DepositHalfEth.test.ts
```diff
@@ -41,43 +39,47 @@ test.describe('Deposit Half ETH', () => {
     await page.waitForTimeout(3000);
     await depositHalfButton.click();
 
-    const walletPage = await getWalletPage(context);
+    const walletNotificationPage =
+      await fuelWalletTestHelper.getWalletNotificationPage();
```
LuizAsFight, line 43: would be better `getWalletPopupPage`?

```suggestion
    const walletPopupPage =
      await fuelWalletTestHelper.getWalletPopupPage();
```

LuizAsFight (changes requested):

luizstacio (approved):

LuizAsFight (approved):

PR #987, opened 2023-11-22
docs: configure spell-check

sarahschwartz:
This PR configures the spell-check added to the docs workflow in the github-actions repo: https://github.com/FuelLabs/github-actions/pull/23

The files checked are configured in `.spellcheck.yml`.  This is also where you can configure what types of elements are ignored.

Right now it ignores:
 - All code blocks that have a language (and will check code blocks that do not have a language)
- Anything in between backticks
- Words in `spell-check-custom-words.txt` (case sensitive, only exact match)
- Numbers, even if they are attached to a word
- Links in markdown link format

#### Handling errors:

If the test fails:
- look up the word in the question to verify it is a real word and is correctly spelled
- If it is a file name or is code, use backticks to ignore the word.
- If it is a real word that is spelled correctly, or an acronym that is either common or is defined already, add it to `spell-check-custom-words.txt`.
- If needed, rewrite the sentence. Ex: DON'T use  "`lock`ing" and add "ing" to the custom words list. Instead, rewrite the sentence as "locking with the `lock` method".
- If it otherwise should be ignored, you can configure the pipeline in `.spellcheck.yml`.


LuizAsFight (approved):
great work here !

@sarahschwartz can the list of words be in the CI itself inside of in this project ? then it can be easily reused across other projects

sarahschwartz:
It can wait as long as you don't make any other changes to the docs, since the docs workflow is broken without this. 

And I thought about that, but I'm afraid that would slow down development since you would have to update 2 repos.

luizstacio:
> It can wait as long as you don't make any other changes to the docs, since the docs workflow is broken without this.
> 
> And I thought about keeping the custom words list in one place, but I'm afraid that would slow down development since you would have to update 2 repos. Most tech terms need to be added to the list.

We could have a default list and allow a per repo bases list also. In this way we would make the CI simpler, and if we need to customize it would be fast.

sarahschwartz:
> We could have a default list and allow a per repo bases list also. In this way we would make the CI simpler, and if we need to customize it would be fast.

That could work, I will follow up with another PR to do this later if that's ok.

luizstacio (approved):

luizstacio (approved):

PR #990, opened 2023-11-27
feat: update beta-5 compatibility

luizstacio:
closes FE-36


packages/app/playwright/e2e/SendTransaction.test.ts
```diff
@@ -122,7 +123,7 @@ test.describe('SendTransaction', () => {
     await hasText(page, 'success');
   });
 
-  test('Send max amount transaction', async () => {
+  test.only('Send max amount transaction', async () => {
```
matt-user, line 126: why are we only running this test now?

packages/app/playwright/commons/gas.ts
```diff
@@ -5,7 +5,7 @@ import type { Provider, BN } from 'fuels';
 export const getGasConfig = async (provider: Provider) => {
   const chain = await provider.getChain();
   const nodeInfo = await provider.fetchNode();
-  const gasLimit = chain.consensusParameters.maxGasPerTx;
+  const gasLimit = chain.consensusParameters.maxGasPerTx.div(2);
```
matt-user, line 8: why are we dividing by 2?

packages/app/src/systems/Send/hooks/useSend.tsx
```diff
@@ -41,7 +41,7 @@ const selectors = {
           txStatus === TxRequestStatus.loading ||
           txStatus === TxRequestStatus.sending;
 
-        if (state.matches('fetchingFakeTx')) return SendStatus.loading;
+        // if (state.matches('fetchingFakeTx')) return SendStatus.loading;
```
matt-user, line 44: can we delete this?

packages/app/playwright/e2e/SendTransaction.test.ts
luizstacio: It should not be like this nice catch

packages/app/playwright/commons/gas.ts
luizstacio: The current state of the fuel-core consider max gas limit, the gas for the entire transaction. But the gasLimit config is used only for script, because of that we need to set the gasLimit in a value that is smaller them the max with room for other costs. This is why we divide by 2.

LuizAsFight (changes requested):

packages/app/src/systems/Send/machines/sendMachine.ts
```diff
@@ -136,19 +120,68 @@ export const sendMachine = createMachine(
...
+                    assetId: BaseAssetId,
+                    amount: nativeBalance,
+                  },
+                ]
+              );
+              transferRequest.addResources(resources);
+              const { gasUsed, gasPrice, usedFee, minFee } =
+                await provider.getTransactionCost(transferRequest);
+              transferRequest.gasPrice = gasPrice;
+              transferRequest.gasLimit = gasUsed;
```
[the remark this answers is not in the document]
luizstacio: I didn't get what we should increase here?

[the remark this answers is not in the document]
luizstacio: I don't think is needed in this case as a transfer gas a very predictable gas usage

LuizAsFight (approved):

PR #996, opened 2023-12-05
style: change text for creating new account in existing wallet

### LINKED ISSUE #992: Text for creating a new account in an existing wallet
This is a minor issue with the messages when creating an account.

We use the word "Add" for both importing and creating a new account.

We should change the text to be "Import from private key" and "Create new account" as that is accurately describing what is happening.

richardgreg:
Changed the text for "Add from private key" and "Add new account" to "Import from private key" and "Create new account."

close #992


packages/app/src/systems/Account/pages/Accounts/Accounts.tsx
```diff
@@ -52,7 +52,7 @@ export const Accounts = () => {
           iconSize={14}
           isLoading={isAddingAccount}
         >
-          Add new account
+          Create new account
```
LuizAsFight, line 55: can you keep this one as it was before?


the "import" change looks good

LuizAsFight (changes requested):

luizstacio:
This PR is missing the changesets. Can you add it by running: `pnpm changeset add --empty`

luizstacio (approved):

PR #1043, opened 2024-01-15
fix: throw error when asset name is undefined

### LINKED ISSUE #935: Showing wrong card when adding asset without "name" prop
image

matt-user:
closes #935


packages/app/src/systems/Asset/services/assets.ts
```diff
@@ -64,6 +64,12 @@ export class AssetService {
 
   static async addAssets(input: AssetInputs['addAssets']) {
     return db.transaction('rw', db.assets, async () => {
+      const someNameUndefined = input.data.some((asset) => {
+        asset.name === undefined;
+      });
+      if (someNameUndefined) {
+        throw new Error('Asset.name is undefined');
+      }
```
luizstacio, line 72: This should be implemented inside the method `validateAddAssets` like the other validations: https://github.com/FuelLabs/fuels-wallet/blob/21540c5ddc921f052f5c5fde040add5b7cf3cd7c/packages/app/src/systems/CRX/background/services/BackgroundService.ts#L328

luizstacio (changes requested):

packages/app/src/systems/Asset/services/assets.ts
```diff
@@ -109,6 +109,14 @@ export class AssetService {
       throw new Error('No assets to add');
     }
 
+    // validate that all of the names are defined
+    const someNameUndefined = assets.some((asset) => {
+      asset.name === undefined;
+    });
```
luizstacio, line 115: I tested using the documentation add asset and the screen opens, I think you need to validate not only undefined but also is is not empty. It's better to also trim the the name to avoid empty spaces counting as a valid name. 

The following values should throw `'', ' ', '       ', undefined, null`

Screenshot 2024-01-15 at 19 14 07

luizstacio (changes requested):

luizstacio (approved):

PR #1045, opened 2024-01-15
fix: check supported network version

### LINKED ISSUE #1036: Add check on add network for checking supported version
As reported here: #1004, we currently don't query the url before adding when the request to add a new network comes via SDK.

To avoid this we should add a new check that queries the version of the network and if not supported returns a error also when adding via SDK.

matt-user:
closes #1036

### DIFF
--- a/packages/app/src/systems/Network/services/network.ts
+++ b/packages/app/src/systems/Network/services/network.ts
@@ -160,10 +160,10 @@ export class NetworkService {
       const provider = await Provider.create(url);
       nodeInfo = await provider.fetchNode();
     } catch (err: unknown) {
-      throw new Error('Node version is not compatible with the wallet.');
+      throw new Error(`Network version is not compatible with >=${MIN_NODE_VERSION} required by the Wallet`);
     }
     if (compare(nodeInfo.nodeVersion, MIN_NODE_VERSION, '<')) {
-      throw new Error('Node version is not compatible with the wallet.');
+        throw new Error(`Network version is not compatible with >=${MIN_NODE_VERSION} required by the Wallet`);
     }
     const collection = await db.transaction('r', db.networks, async () => {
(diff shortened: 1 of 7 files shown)


packages/app/src/systems/Network/services/network.ts
```diff
@@ -148,11 +149,22 @@ export class NetworkService {
...
+    try {
+      const provider = await Provider.create(url);
+      nodeInfo = await provider.fetchNode();
+    } catch (err: unknown) {
+      throw new Error('Node version is not compatible with the wallet.');
+    }
+    const nodeVersion = parseInt(nodeInfo.nodeVersion.replace('.', ''));
+    if (nodeVersion < MIN_NODE_VERSION) {
+      throw new Error('Node version is not compatible with the wallet.');
+    }
```
luizstacio, line 167: We can use `compare-versions` to compare here, them we can have only the string on the config also, `0.22.0`.

luizstacio (changes requested):

packages/app/src/systems/Network/services/network.ts
```diff
@@ -148,11 +150,21 @@ export class NetworkService {
...
     const { name, url } = input.data;
     if (!isValidNetworkUrl(url)) {
       throw new Error('Invalid network URL');
     }
+    let nodeInfo;
+    try {
+      const provider = await Provider.create(url);
+      nodeInfo = await provider.fetchNode();
+    } catch (err: unknown) {
+      throw new Error('Node version is not compatible with the wallet.');
```
luizstacio, line 163:
```suggestion
      throw new Error(`Network version is not compatible with >=${MIN_NODE_VERSION} required by the Wallet`);
```

luizstacio (approved):

PR #1063, opened 2024-01-23
feat: show broken tx

matt-user:
closes FE-78


matt-user:
one way to test this pr

1. modify a call in e2e-contracts custom interactions such that it will fail (gas limit)
2. deploy the e2e contracts
3. dev:e2e-contract-tests
4. call the contract

luizstacio (changes requested):
- When setting the gasLimit to a small value I got this error, it should show the the message as error but shows `Carefully check if all the details in your transaction are correct. Operations:`.
- Also we could remove the icon to center better the error message.
Screenshot 2024-01-25 at 23 04 56

luizstacio (approved):

Issue #1095, opened 2024-02-20
Long account name in wallet extension can block access to wallet


natszpre:
### What version of Wallet are you using?

0.13.10

### What is happening?

Are there any rules what the wallet account name can contain? I have just saved very long name, and I'm now not able to set it back because the button is probably somewhere in the middle, and I'm not able to scroll to that point. Also, all the other buttons just disappeared, amount on the account is not visible. The whole extension view is occupied with the account name, and I'm not able to use it anymore.

### steps to reproduce

Change the account name to very long string and save

### How it should behave?

It should be a limit on the name length, all the buttons should be still visible (it should be possible to scroll to the bottom of the page)
![Screenshot from 2024-02-20 13-55-56](https://github.com/FuelLabs/fuels-wallet/assets/5881837/8324112c-125c-4682-864c-a82a9ce72d2d)

### What os are you seeing the problem on?

None

### What browser?

Chrome

### What is your browser version?

Version 122.0.6261.39 (Official Build) Arch Linux (64-bit)

luizstacio:
@natszpre thanks for the issue nice catch here.

PR #1118, opened 2024-02-28
fix: display truncated account name correctly

helciofranco:
It fixes https://github.com/FuelLabs/fuels-wallet/issues/1095

| Before | After |
|-------------|-------------|
| Screenshot 2024-02-28 at 15 01 31 | Screenshot 2024-02-28 at 15 01 19 |

### DIFF
--- a/.changeset/tender-mails-float.md
+++ b/.changeset/tender-mails-float.md
@@ -1,0 +1,5 @@
+---
+'fuels-wallet': patch
+---
+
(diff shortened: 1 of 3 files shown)


pedronauck (approved):

luizstacio (changes requested):
We should fix on the account dropdown also, I think we could add a max length to the account name.
Screenshot 2024-02-28 at 18 15 47

PR #1123, opened 2024-03-01
perf: eliminating asset avatar flashing

bethatguyad:
This pull request memo the `Avatar` component, which avoids unnecessary rerendering when `imageUrl` is unchanged.
## Before

[before.webm](https://github.com/FuelLabs/fuels-wallet/assets/160697994/05275c41-fc8d-4ea2-ac09-186049b64203)

## After

[after.webm](https://github.com/FuelLabs/fuels-wallet/assets/160697994/0af2c0d0-7c74-4f6f-9a79-c3a2ca40b4ea)

### DIFF
--- a/packages/app/src/systems/Asset/components/AssetItem/AssetItem.tsx
+++ b/packages/app/src/systems/Asset/components/AssetItem/AssetItem.tsx
@@ -13,7 +13,7 @@ import {
 } from '@fuel-ui/react';
 import type { AssetAmount } from '@fuel-wallet/types';
 import { bn } from 'fuels';
-import { useMemo, type FC } from 'react';
+import type { FC } from 'react';
 import { useNavigate } from 'react-router-dom';
 import { AmountVisibility, Pages, shortAddress } from '~/systems/Core';
 import { useBalanceVisibility } from '~/systems/Core/hooks/useVisibility';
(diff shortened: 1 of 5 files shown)


bethatguyad:
@helciofranco friendly ping, would you mind having a look?

helciofranco:
Hi @bethatguyad Thanks for you contribution. Could you please add a `changeset` for this change?

```pnpm changeset```

**Comment by bethatguyad** [comment]
> Hi @bethatguyad Thanks for your contribution. Could you please add a `changeset` for this change?
> 
> `pnpm changeset`

Done

**Comment by bethatguyad** [comment]
It looks like the test failure is caused by this PR, you could find it other where too, 
image
https://github.com/FuelLabs/fuels-wallet/pull/1121

**Comment by bethatguyad** [comment]
Hi, folks is there any change to merge this pull request?

File: packages/app/src/systems/Asset/components/AssetItem/AssetItem.tsx [hunk 1]
```diff
@@ -108,17 +108,22 @@ export const AssetItem: AssetItemComponent = ({
...
+    return imageUrl ? (
+      <Avatar
+        name={name}
+        src={imageUrl}
+        css={{ height: 36, width: 36, borderRadius: '$full' }}
+      />
+    ) : (
+      <Avatar.Generated hash={assetId} css={{ height: 36, width: 36 }} />
+    );
+  }, [name, imageUrl, assetId]);
```
helciofranco, line 122: I've just checked it deeper and noticed not only these two componentes are re-rendering.

Screenshot 2024-03-13 at 09 45 30

The issue is that `AssetList` is being re-render due to its parent re-render.
And, because of that, `AssetItem` as well.


Could you revert this change and wrap `AssetList` with a `memo`?

e.g.
```tsx
export const AssetList = memo(AssetListBase);
```

With this change I believe you'll also need to replace `AssetList.Loading` with `AssetListLoading`, for example.

helciofranco (changes requested):

packages/app/src/systems/Asset/components/AssetItem/AssetItem.tsx
helciofranco: I forgot one small detail, to make it works, we'll need also to memoize an object on `Home`

https://github.com/FuelLabs/fuels-wallet/blob/7dde82f84a8397f79485607886e4b362f0f22698/packages/app/src/systems/Home/pages/Home/Home.tsx#L51-L53

```tsx
const emptyProps = useMemo(() => {
  return {
    showFaucet: selectedNetwork?.url === VITE_FUEL_PROVIDER_URL,
  };
}, [selectedNetwork]);
```

helciofranco (approved):

PR #1155, opened 2024-03-19
feat: updating sdk to latest

fuel-service-user:
✨ This PR updates the SDK to tag latest


helciofranco:
It's working ✅ 
I've requested update of `react-hook-forms` just to check how it would work for other packages.

Screenshot 2024-03-19 at 18 57 43

PR #1180, opened 2024-03-22
(WIP) Bako ID integration

luisburigo:
This feature enables sending a transaction by handles in Bako, handle is `@` with a identifier name, an example:
`@fuel_network` 👇
```json 
{
  "owner": "0x000...",
  "resolver": "0x000...",
}
```

[Check Bako ID docs](https://docs.bako.id/)

---

Changes:
- [x] Modify address input to accept a `Bech32` or `@bako_handle`
- [x] Validations on input address
- [x] Fetch Bako handle and use the `resolver` address
- [x] Include `@bako-id/sdk@0.0.3` package
- [ ] Show handle name In transaction approve page  
- [ ] Create tests cases

### DIFF
--- a/packages/app/package.json
+++ b/packages/app/package.json
@@ -20,6 +20,7 @@
     "codegen": "graphql-codegen --config codegen.ts"
   },
   "dependencies": {
+    "@bako-id/sdk": "0.0.3",
     "@fontsource/source-code-pro": "5.0.13",
     "@fuel-ui/css": "0.23.2",
     "@fuel-ui/icons": "0.23.2",
--- a/packages/app/src/systems/Send/components/SendSelect/SendSelect.tsx
+++ b/packages/app/src/systems/Send/components/SendSelect/SendSelect.tsx
@@ -1,10 +1,15 @@
 import { cssObj } from '@fuel-ui/css';
-import { Box, Input, InputAmount, Text } from '@fuel-ui/react';
+import { Box, Form, Input, InputAmount, Text } from '@fuel-ui/react';
 import { motion } from 'framer-motion';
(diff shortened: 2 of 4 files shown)


packages/app/src/systems/Transaction/components/TxRecipientCard/TxRecipientCard.tsx
```diff
@@ -29,8 +40,41 @@ export const TxRecipientCard: TxRecipientCardComponent = ({
...
+      setAccountName();
+    };
+
+    if (!name && isReceiver) {
+      resolveAddressName();
+    }
+
+    if (!isReceiver) {
+      setAccountName();
+    }
```
luisburigo, line 68: I added it only to show in the receiver, thinking about the transaction sending flow, what do you think?

If they're not a receiver, there could be a conflict with the wallet name.

packages/app/src/systems/Transaction/components/TxRecipientCard/TxRecipientCard.tsx
```diff
@@ -29,8 +40,41 @@ export const TxRecipientCard: TxRecipientCardComponent = ({
...
+  }, [isReceiver, name]);
+
+  const Name = !name ? (
+    <ContentLoader width={100} height={17} viewBox="0 0 80 17">
+      <ContentLoader.Rect x="0" y="0" width="100" height="17" rx="4" />
+    </ContentLoader>
+  ) : (
+    `${name}`
+  );
```
luisburigo, line 77: I added a loader to the name to avoid screen flickering when the search is completed.

PR #1182, opened 2024-03-22
fix: turn off autocomplete on the private key input

helciofranco:
- Disable `autoComplete` for the Private Key field (security reason).
- Make it `password` type as well.


HoTandy (approved):
Installed the latest dev extension and autocomplete is no longer happening on this field. Perfect

PR #1186, opened 2024-03-25
chore: replace warning icon with a info circle icon on connected apps list

helciofranco:
- Fixes tooltip title text alignment.
- Replaces that orange `warning icon` with a gray `info circle icon`.

| ⚠️  Before |  ℹ️ After |
| ---- | ---- |
| before | aftert |

### DIFF
--- a/examples/cra-dapp/package.json
+++ b/examples/cra-dapp/package.json
@@ -1,5 +1,6 @@
 {
   "name": "cra-dapp",
+  "type": "module",
   "version": "0.1.0",
   "private": true,
   "scripts": {
--- a/.changeset/short-feet-provide.md
+++ b/.changeset/short-feet-provide.md
@@ -1,0 +1,5 @@
+---
+"fuels-wallet": patch
+---
+
+chore: replaces a warning icon (orange) with a info icon (gray) from the connection list screen.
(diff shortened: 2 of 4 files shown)


HoTandy (approved):
The new grey information icon looks far less dramatic and shouldn't scare the user. Looks good.

examples/cra-dapp/package.json
```diff
@@ -1,5 +1,6 @@
 {
   "name": "cra-dapp",
+  "type": "module",
```
LuizAsFight, line 3: we need this change ?

LuizAsFight: looks orthogonal to this PR

LuizAsFight (changes requested):

examples/cra-dapp/package.json
helciofranco: We don't really need it, but it fixes a warning when running our example dapp.

Screenshot 2024-03-29 at 19 40 09

That link recommends that we set up ESM as default in our `package.json`.
Reference: https://vitejs.dev/guide/troubleshooting.html#vite-cjs-node-api-deprecated

If you don't mind we can move forward with this change.

LuizAsFight (approved):

PR #1195, opened 2024-03-26
fix: disable autocomplete everywhere

helciofranco:
@HoTandy found more inputs autocompleting automatically. By the way, autocomplete is always a tricky thing to turn off.
Now I'm implementing a different strategy by prefixing a few inputs with `search-` id. 

It makes even passwords apps like `1Password` ignore these fields. Which is also a kind of autocomplete but not handled natively by the browser.
 
Check it out:

### Asset form (1password or any other vault app)
| Before | After |
| --- | --- |
| 1password | Screenshot 2024-03-26 at 14 21 54 |

### Sending transactions (chrome autocomplete)
| Before | After |
| --- | --- |
| before | Screenshot 2024-03-26 at 14 22 48 |

### Updating password (chrome autocomplete)
| Before | After |
| --- | --- |
| Screenshot 2024-03-26 at 14 17 28 | Screenshot 2024-03-26 at 14 24 47 |


HoTandy (commented):
The 3 different cases you've provided look good when testing them, however, as I was flying around the wallet, I saw more cases of this. Below are the screenshots:

image

and

image

and

image

I don't know if you want to address these as well before I approve, ...so I'll just comment for now.

helciofranco:
@HoTandy nice catch, I'll address those findings on this PR too.

LuizAsFight (changes requested):
only waiting on @HoTandy review, but code-wise is good to go

HoTandy:
Clicking on the last View Deployment button above, then installing the wallet from that vercel page (regardless of whether I install from the zip or from Chrome), I'm still seeing autocomplete on some fields. For example:

image

Am I verifying correctly?

helciofranco:
@HoTandy yes, that’s correct, I haven’t fixed your latest finds yet. I will try to find a time during this week to finish this.

helciofranco:
Hey @HoTandy, I've just pushed those missing founds.
Now it's ready for review again.

LuizAsFight (approved):

PR #1196, opened 2024-03-26
fix: add transaction id to the transaction approve page

helciofranco:
When the user is approving a transaction, the `id` is defined only on the `ctx.approvedTx`.
We can't say the same for the `ctx.txResult`, which has an undefined `id` in this flow.

I've updated to use that `id` when rendering the `TxApprove` screen.
Now the user can click to open the explorer properly.


HoTandy (approved):
Perfect

luizstacio (approved):

PR #1197, opened 2024-03-26
feat: sync UI between connected apps list and edit screens

helciofranco:
- Display the full app URL when it's big (instead of truncating it). Allowing the user to truly verify their connected apps.
- Sync the information displayed in the connected app `list` and `edit` screens to enhance consistency across our UI.
- Improve how the URL are displayed, focusing on the `hostname` mostly and use a native API instead of regex.
- Adjust the `Avatar` display to show only the first two initials letter when the page title is long (2+ words), preventing UI breakage.

|  List | Edit |
| --- | --- |
| list | Screenshot 2024-03-26 at 17 22 40 |

### DIFF
--- a/.changeset/rich-pens-wave.md
+++ b/.changeset/rich-pens-wave.md
@@ -1,0 +1,8 @@
+---
+"fuels-wallet": patch
+---
+
+- Display the full app URL when it's big (instead of truncating it). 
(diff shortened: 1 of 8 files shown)


HoTandy (approved):
I don't see a View deployment link for this one but the wrapping of the URL looks good!

luizstacio (approved):

PR #1217, opened 2024-04-08
fix: Change Password dialog allowing new password to be the same as current #1203

### LINKED ISSUE #1203: Change password dialog in wallet - multiple issues

Version: 0.16.1

### What is happening?

I see some problems on the Change Password screen.

1. If I choose the same password, the little padlock goes red but I can still put this password into the Confirm Password field and Save the "new" password.
2. I can also get rid of this red padlock by changing the 2 passwords to a different one than the original one, but then changing back 
3. Also, it's possibly to paste into the Confirm Password field. You can't do this when restoring a wallet just after installing the extension, so why is it possible in the wallet?

Image

### steps to reproduce

See above

### How it should behave?

1. Should not be allowed to choose the same password
2. Fix inconsistant behaviour with the red padlock
3. Prevent the user from pasting the confirmation of the password

Mac/Apple Desktop

### What browser?

Chrome

Version 123.0.6312.59 (Official Build) (x86_64)

arthurgeron:
Fixes #1203 
- Prevents change password dialog from allowing the new password from being the same as the current password

| Before     | After |
| ---      | ---       |
| Screenshot 2024-04-08 at 20 27 43 | Screenshot 2024-04-08 at 20 26 12 |

### DIFF
--- a/.changeset/lovely-balloons-behave.md
+++ b/.changeset/lovely-balloons-behave.md
@@ -1,0 +1,5 @@
+---
+"fuels-wallet": patch
+---
+
+Prevent the Change Password dialog from allowing the new password to be the same as the current password #1203
--- a/packages/app/package.json
+++ b/packages/app/package.json
@@ -118,7 +118,7 @@
     "ts-jest-mock-import-meta": "1.1.0",
     "tsconfig-paths-webpack-plugin": "4.1.0",
     "typescript": "5.2.2",
-    "vite": "4.5.2",
+    "vite": "4.5.3",
     "vite-plugin-clean": "1.0.0",
     "vite-plugin-static-copy": "0.17.0",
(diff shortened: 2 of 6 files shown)


HoTandy:
A quick comment before I approve, the change looks good!

I can immediately do a third change to change my password back to the first password after changing it to a different second password. Is this OK?

If so then I will approve.

arthurgeron:
> A quick comment before I approve, the change looks good!
> 
> I can immediately do a third change to change my password back to the first password after changing it to a different second password. Is this OK?
> 
> If so then I will approve.

Yes, that will work! It doesn't check with your actual current password, instead it only compares with what is in the field for `Current Password`, so changing a third time to your original password should work just fine.

HoTandy (approved):

PR #1218, opened 2024-04-09
fix: transaction approval screen showing odd half loading state

### LINKED ISSUE #1201: Wallet in strange state after clicking Approve

Version: 0.16.1

### What is happening?

When a transaction is `Approving`, I can sometimes still click on the `Approve` and `Back` buttons, see image below.

Image

And then if I click on the `Back` button, it goes back to the `Send` screen, but I can no longer click on the `Confirm` button.

Image

This seems to be a strange state that the wallet got itself into, it's difficult to reproduce, I think it's when the fuel network is a bit slow or something.

### steps to reproduce

Actually it's quite difficult to reproduce, it might have something to do with the system being slow or something in the background

### How it should behave?

We should not be able to do this.

Mac/Apple Desktop

### What browser?

Chrome

Version 123.0.6312.59 (Official Build) (x86_64)

arthurgeron:
Fixes #1201  
- Fixes loading state only being partially applied to TxApprove components
- Adds unit tests for TxApprove (as a dumb component)

## Before
Screenshot 2024-04-08 at 20 27 43

## After

https://github.com/FuelLabs/fuels-wallet/assets/3487334/60518df0-3b84-4d4c-881b-cc26dfc9b081

### DIFF
--- a/.changeset/curly-lamps-work.md
+++ b/.changeset/curly-lamps-work.md
@@ -1,0 +1,5 @@
+---
+"fuels-wallet": patch
+---
+
+Fixes Approve Transaction screen staying in a partially loading state after approving a transaction
--- a/packages/app/src/systems/Transaction/pages/TxApprove/TxApprove.test.tsx
+++ b/packages/app/src/systems/Transaction/pages/TxApprove/TxApprove.test.tsx
@@ -1,16 +1,15 @@
 import { fireEvent, render, screen } from '@testing-library/react';
 import { TransactionStatus } from 'fuels';
-import { useAssets } from '~/systems/Asset';
-import { TxRequestStatus, useTransactionRequest } from '~/systems/DApp';
+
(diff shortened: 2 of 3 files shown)


HoTandy:
Hi @arthurgeron, I would like to verify this fix, on which environment should I be verifying it? (vercel*)

arthurgeron:
> Hi @arthurgeron, I would like to verify this fix, on which environment should I be verifying it? (vercel*)

Hello @HoTandy ! I'd recommend the test environment, but Preview should work too. Preferably the environment you were using when you first reproduced this issue!

HoTandy (approved):
Hi @arthurgeron, I've now retested where you deployed https://fuels-wallet-5xb0l6yz9-fuel-labs.vercel.app/docs/install/ - installed from zip on win11.

Looks good, I cannot reproduce the issue anymore, but it was difficult to reproduce in the first place.

helciofranco (approved):

LuizAsFight (approved):

PR #1230, opened 2024-04-12
fix: DB restart should reset service worker communication protocol and services

### LINKED ISSUE #1124: Explore Wallet error around DBIndexed (sentry)

arthurgeron:
Fixes #1124

Previously, when the DB fails, closes or gets blocked:
- Service workers would not react, acting as if the connection was up and healthy.
- Vault would have it's secret lost in the dropped IndexDB but still be locked, leading to a soft-lock state where there's no Wallet setup and it's unable to unlock itself to set/recover a new wallet again

Now when IndexDB drops/fails :
- DB will detect that and restart
- DB will send a "restarted" event
- Event will be detected by Vault's service in service worker, which will re-instantiate it's connection to IndexDB
- The welcome page (if open) will detect the event, and create a internal feature flag
- The extension will reload itself, now with a fresh Vault, this restart will cause all extension related windows to close
- Upon the restart, if the flag previously created is detected, the extension will re-open the welcome window

https://github.com/FuelLabs/fuels-wallet/assets/3487334/25ac310b-3e92-46b8-aeaa-9243adb3c462


LuizAsFight (changes requested):
tests are breaking


also I got this error in console in `import seed` screen
image

HoTandy (commented):
It can be that if I click "clear all site data" more than once, sometimes the wallet will be created after clicking `Next: Finish set-up` and sometimes it's still broken.  I've attached a log from the console from one of the times where it was still broken.

[oiclkmaecneaoeejodlfdggapjpgkeoh-1713432547640.log](https://github.com/FuelLabs/fuels-wallet/files/15022811/oiclkmaecneaoeejodlfdggapjpgkeoh-1713432547640.log)

image

Dev wallet taken from `View Deployment` link https://fuels-wallet-675si7fzx-fuel-labs.vercel.app/docs/install/

Please let me know if you need more info.

arthurgeron:
@LuizAsFight I've figured out what's really causing this bug.
When you "clear all data" after having a wallet created/imported, you drop the IndexedDB with the secret in it. The Vault will remain locked, but unable to retrieve the secret, which leads to a soft-lock state where it sends you to the welcome page but is unable to do anything (clear, add wallets/accounts, etc).

I've rolled back most of my changes, since they were overkill in comparison to the fix that I've come up with.
Now Vault listens to `DB Restarted` events, then executes a integrity check, which basically:
- Checks if DB is open/reachable
- Checks if the vault is unlocked (which is the "OK" scenario)
If the DB is unreachable/dropped, or the vault is locked but the secret cannot be retrieved, that will trigger the re-instantiation of the IndexedDB and a complete reload of the extension, which will prevent the wallet from entering a soft locked state.

@HoTandy  Please retest when possible! This PR is now ready to merge.

HoTandy (commented):
Hi @arthurgeron, now the wallet creation is working as expected after clearing all site data and then recreating the wallet.

Interestingly though, clearing all site data is killing the tab completely. Should this happen?  If the closing of the tab is classed by you as a none issue that I'm happy to Approve the change.

https://github.com/FuelLabs/fuels-wallet/assets/114662397/68d00f05-2df3-4dfd-b3e9-21e3d3e127b5

arthurgeron:
@LuizAsFight  and @luizstacio @pedronauck  and @helciofranco  please let me know your thoughts on the implementation mentioned in my previous comment.

HoTandy (approved):
This is working very nicely now with the page reload.

packages/app/src/hooks.ts
```diff
@@ -0,0 +1,28 @@
...
+    };
+
+    if (IS_CRX) {
+      chrome.runtime.onMessage.addListener(handleRestartEvent);
+    }
+
+    return () => {
+      chrome.runtime.onMessage.removeListener(handleRestartEvent);
+    };
+  }, []);
```
LuizAsFight, line 27: makes sense to move this hook to inside of `systems/Core/hooks`?

LuizAsFight (changes requested):

packages/app/src/hooks.ts
arthurgeron: You're right, I've fixed it! Thank you

LuizAsFight (approved):

PR #1243, opened 2024-04-23
feat: add auto update wallet

luizstacio:
With this PR, the wallet will now check if it is running an old version and will auto-update once a new version is available on the background, respecting the user flow, if the users had any Wallet popup / tab open it will not update the wallet and delay it for more 15 minutes.


arthurgeron (dismissed):

packages/app/env.d.ts
```diff
@@ -7,6 +7,7 @@ declare namespace NodeJS {
     readonly VITE_MNEMONIC_WORDS: string;
     readonly VITE_ADDR_OWNER: string;
     readonly VITE_CRX_NAME: string;
+    readonly VITE_API_VERSION: string;
```
LuizAsFight, line 10:
```suggestion
    readonly VITE_CRX_VERSION_API: string;
```

LuizAsFight: or remove it?

LuizAsFight (changes requested):

LuizAsFight (approved):

PR #1273, opened 2024-05-06
chore: remove recaptcha

helciofranco:
Removing `reCAPTCHA` from the wallet extension.


packages/app/playwright/e2e/HomeWallet.test.ts
```diff
@@ -21,20 +21,6 @@ test.describe('HomeWallet', () => {
     await mockData(page);
   });
 
-  test('should change balance when select a new network', async () => {
```
luizstacio, line 24: You probably need another way to fund the wallet here.

luizstacio (approved):

PR #1277, opened 2024-05-06
feat: add advanced fees when approving a tx request

helciofranco:
This PR extends the `Custom Fees` feature from the `Send Transaction` screen, enabling users to set a **custom fee** when approving a request from a Dapp.

| 📷 Demo |
| --- |
| <video src="https://github.com/user-attachments/assets/7171f53b-d341-454f-b2d8-0a5b8702e328" /> |


LuizAsFight (changes requested):
when I change `gasLimit` to a really high value, the fee is not changing at all.. only the `tip` is changing the fee, can you check if the transaction is being estimated with the inputted `gasLimit` ? this doesn't stop this PR but we should detect if the problem is in the wallet or in the ts sdk

image

LuizAsFight (approved):

LuizAsFight (approved):

PR #1279, opened 2024-05-06
fix: vault crash on export seed auth fail

arthurgeron:
Fixes soft-locked state after wrong password in `View Seed Phrase` screen


packages/app/src/systems/Settings/hooks/useExportVault.tsx
```diff
@@ -20,15 +20,19 @@ export function useExportVault() {
...
-      },
-    });
+    vault &&
+      service.send({
+        type: 'EXPORT_VAULT',
+        input: {
+          password,
+          vaultId: vault.vaultId,
+        },
+      });
```
luizstacio, line 35: We should not access the vault directly on the UI components. We should move the logic to get vaults method.

luizstacio (changes requested):

HoTandy (approved):
Apologies for the late review but looks good!!

LuizAsFight (approved):

PR #1280, opened 2024-05-06
fix: vault server lock propagation

arthurgeron:
## Context
`useUnlock` listens for `lock` and `unlock` events from VaultClient, but the communication with VaultServer was one-way, meaning those events from the Server never reached the client. That means that a lock changes from server-side would never be reflected in our UI.
In that scenario the wallet locks but the user wouldn't be redirected to the auth screen, being able to keep interacting with UI elements behind the auth guard, leading to several bugs.

## Solution
Fixed by using propagating VaultServer lock change events to VaultClient through existing communication protocol, VaultClient further propagates those events, which are now correctly handled by `useUnlock`, leading to proper behavior (e.g. redirection to the auth screen).

### Before   
https://github.com/FuelLabs/fuels-wallet/assets/3487334/c85944cc-fe7e-4a62-a670-4414d790dc98
### After   
https://github.com/FuelLabs/fuels-wallet/assets/3487334/1de77275-b18d-474f-b3bb-e32153832d2e


packages/app/src/systems/Vault/services/VaultServer.ts
```diff
@@ -80,6 +80,19 @@ export class VaultServer extends EventEmitter {
       }
       this.server.addMethod(methodName, this[methodName].bind(this));
     });
+
+    this.manager.on('unlock', () => {
+      chrome.runtime.sendMessage({
```
luizstacio, line 85: You can use this.server.notify to send messages from here to the client. https://www.npmjs.com/package/json-rpc-2.0

This is better to ensure the events are submitted in a standard way, avoid the use directly of the chrome.runtime

packages/app/src/systems/Vault/services/VaultClient.ts
```diff
@@ -28,6 +33,14 @@ export class VaultClient extends EventEmitter {
     VaultServer.methods.forEach((methodName) => {
       this[methodName] = this.createRequest(methodName);
     });
+
+    if (IS_CRX) {
+      chrome.runtime.onMessage.addListener((e: LockChangeEvent) => {
```
luizstacio, line 38: With the implementation on the server of notify you need to implement [here](https://github.com/FuelLabs/fuels-wallet/blob/0248a70099d9202043a825da917ed2a4673311aa/packages/app/src/systems/Vault/connectors/VaultCRXConnector.ts#L27) and here to receive the correct type.

luizstacio (changes requested):

packages/app/src/systems/Vault/services/VaultServer.ts
arthurgeron: @luizstacio 
Thank you for the suggestion to use this.server.notify for event submissions. However, it's important to note that JSON-RPC 2.0, while supporting structured message handling, does not inherently manage the transport layer. To enable this.server.notify to send messages from the server to the client, we would need a [bi-directional communication setup](https://www.npmjs.com/package/json-rpc-2.0#bi-directional-json-rpc), such as a custom WebSocket server and client. This setup would handle the transport of JSON-RPC messages bi-directionally, but would require significant development and maintenance efforts.

Given the complexity of implementing a custom WebSocket infrastructure, should we proceed with this approach considering the potential overhead?

luizstacio (approved):

HoTandy (changes requested):
Hi Arthur, in your "After" example in the PR, the little popup `Invalid Credentials` pops up for a few seconds, with me this isn't happening, therefore there's no feedback to the user.  Could you please take a look. Thanks!

https://github.com/FuelLabs/fuels-wallet/assets/114662397/4a880aa4-71fb-4566-8486-f9937dab1888

HoTandy (approved):

PR #1281, opened 2024-05-07
fix: for starting accounts at Account 1 and added new test

HoTandy:
- Using the mock object, the Accounts were labelled starting at `Account 0`, this doesn't match reality so now they start at `Account 1`

- In popup.ts, I've added functions for doing the following through the UI
-- creating an account - `createAccount`
-- creating an account from private key - `createAccountFromPrivateKey`
-- extracting text from the clipboard - `getClipboardText`
-- getting the fuel address for an account number - `getAddressForAccountNumber`
-- exporting the private key - `exportPrivateKey` (not yet used and not sure if it's working properly)

In `Accounts.test.ts` I've created a test that checks to make sure the same addresses are getting added when we add new accounts, even after importing an account using that account's private key.


helciofranco (commented):
Just two comments:

1. Missing to add a changeset for these changes. 
It can be an "empty" changeset since these changes are not in the app itself.
```pnpm changeset --empty``` and then commit and push that file.

2. Also, the PR title should be updated to use the prefix "`fix:`" (there's a CI routine failing due to the PR title).

helciofranco (approved):

LuizAsFight (changes requested):
why pnpm-lock changed? I dont see any changes in dependencies

LuizAsFight (approved):

PR #1288, opened 2024-05-13
fix: getting started routing not working on other path structures

arthurgeron:
- Fixes routing for "Getting Started" not working in docker hub
 
 
 ## Context:
 Wallet's Docs use a base path for all the routes, so if the base route is `http://localhost:3030/docs/dev/`, then `Getting Started`page is going to be `http://localhost:3030/docs/dev/getting-started/`. 

## The problem:
 The problem with the current approach is that it'd push the entire path, so to change from `getting started` to install it'd need to push `/docs/dev/install/`, which is problematic because it isn't agnostic to the base path, which will change with implementations such as the one in the docs hub.
 
 ## Solution:
 Assume the base path from the current path in install page, so for `http://localhost:3030/docs/dev/install/` it'd calculate the base path to be `http://localhost:3030/docs/dev/`, appending the new Wallet Docs path to the end of that base, which essentially makes it agnostic regarding the base path, fixing the behavior of that route navigation in Docs Hub.


packages/docs/src/components/SDKSection.tsx
```diff
@@ -1,11 +1,19 @@
...
 import connectorsPreviewImg from '../../public/connectors-preview.png';
 
 export function SDKSection() {
   const { push } = useRouter();
+  const pathname = usePathname();
+
+  // Allows routing to work regardless of path structure (e.g. docs hub)
+  function routeToGettingStarted() {
+    const basePath = pathname.split('/').slice(0, -2).join('/');
+    const newCompletePath = `${basePath}/dev/getting-started/`;
```
helciofranco, line 14: I think to keep it really stable we might get it from `window.location.host`
otherwise it will break if we change the URL structure for something different from "2" level of folders

LuizAsFight (approved):

PR #1290, opened 2024-05-13
fix: flaky e2e tests

### LINKED ISSUE #1262: FuelWallet Extension › SDK operations - e2e test failing
The following e2e test is failing on my machine.

`FuelWallet Extension › SDK operations`

Image

Steps to reproduce.

1. Clone `fuels-wallet` repo
2. `pnpm install`
3. `pnpm exec playwright install --with-deps chromium`
4. `cp packages/app/.env.example packages/app/.env` (to setup our network)
5. `pnpm node:up` (to start a local fuel node)
6. `pnpm dev:crx` (to start the extension locally)
7. In a new terminal, run `pnpm test:e2e`

Expected:

All e2e tests should pass

Additionally, the test also fails in the IDE (not just from the command line):

image

ARM based:
ProductName:  `macOS`
ProductVersion:  `14.4.1`
BuildVersion:  `23E224`
Chrome: Version 124.0.6367.78 (Official Build) (arm64)

### LINKED ISSUE #1298: Use new `fuels-npm-packs` which supports devnet
after doing it, make sure that:
- e2e tests and e2e-contract-tests are running smooth both on CI and local
- e2e-contract-tests UI, all contract calls should be working fine locally

arthurgeron:
Closes #1298 
Closes #1262 
- Re-enabled E2E contract tests
- Fixed biome trying to validate files generated by playwright test runs
- Fixed Playwright tests starting up before CRX files are available, now runs `build` beforehand.
- Fixed CRX dev server reloading during playwright tests, causing instabilities
- Fixed NODE_ENV being set to default (production) on local E2E tests, leading to the wrong version (i.e. production) being installed
- Fixed bug where only the first test file in a suite run would use the local CRX build, it'd then download and use a production build of the Wallet for the rest


helciofranco (approved):

LuizAsFight (changes requested):
run the `delply:contracts` locally is breaking

arthurgeron:
> run the `delply:contracts` locally is breaking

Good catch. That was caused by Biome trying to parse `playwright-report` files, it'd only happen if you try to re-deploy the contracts after running e2e tests once. Now fixed!

package.json
```diff
@@ -50,8 +50,8 @@
     "test:ci": "turbo run test --parallel -- --ci --testLocationInResults --json --coverage",
     "test:clear": "pnpm -r exec jest --clearCache",
     "test:coverage": "turbo run test --parallel -- --coverage",
-    "test:e2e": "playwright test --config=packages/app/playwright.config.ts",
-    "test:e2e:contracts": "playwright test --config=packages/e2e-contract-tests/playwright.config.ts",
+    "test:e2e": "pnpm build:crx && playwright test --config=packages/app/playwright.config.ts",
```
LuizAsFight, line 53: you can use `run-s` here as we have it setup in the project

LuizAsFight (changes requested):

package.json
arthurgeron: Fixed

arthurgeron: That's true, thanks! It's fixed now

LuizAsFight (approved):

PR #1319, opened 2024-05-28
feat: validate asset names during e2e contract tests

arthurgeron:
- Added aria labels to Recipient Sender and Asset Name on transaction approval screen;
- Create util for checking text content on multiple aria labels;
- Validate all asset names during e2e contract tests on the transaction approval screen, executes validation on all e2e contract test files.


packages/app/src/systems/Asset/components/AssetsAmount/AssetsAmount.tsx
```diff
@@ -43,7 +43,7 @@ export const AssetsAmount: AssetsAmountComponent = ({
             <Box.Flex css={styles.header}>
               {title && (
                 <Text as="h3" css={styles.title}>
-                  {title}
+                  {title}ss
```
LuizAsFight, line 46: leftover?

LuizAsFight (changes requested):

packages/app/src/systems/Asset/components/AssetsAmount/AssetsAmount.tsx
arthurgeron: Yes! Good catch, thanks

LuizAsFight (approved):

PR #1420, opened 2024-07-31
refactor: sentry integration

### LINKED ISSUE #1404: Investigate sentry bugs
next round of iteration on investigate reported errors on sentry:
- fix the ones we identify
- try to fix the link of sourcemap with the error, right now looks like it's not accurate

arthurgeron:
Closes #1404 

# Changes

- Refactored Sentry Implementation:
  - Use single global instantiation, fixes previous global state pollution
  - Sentry now uses recommended React integrations (e.g. react dom router)
- Error handling logic centralized in Report Error Machine
- Errors are now asynchronously fed into our error Indexed DB from every source (i.e. Service Worker, setup page, extension)
- Error Machine watches errors indexedDB for changes and automatically updates  its own state
- Extension no longer completely reloads when dismissing/ignoring errors, now gracefully recovers from the error state.
- Audit check now ignores vulnerabilities with no known patched version
- Fixed broken links to `@fuels/react` in docs after it migrated to Connectors

# Features
- New error review screen
  - User can dismiss individual errors and/or remove non essential fields from errors before sending them to Sentry
  - User can review the Error's contents and identify if any private information is being sent
  - Error data is automatically sanitized for keys and other obviously private information
  - Aside from protected/required properties (i.e. message & stack), other properties can be deleted by the user to avoid leaking private information

# Evidence
## Error Review Screen
CleanShot 2024-08-15 at 18 57 14

## Error Review Flow (Outdated) - Required error properties protection disabled:
https://github.com/user-attachments/assets/d54f80b4-3da4-4aa8-be0d-60399c1d92b8

## Example Sentry error with proper stack trace and source mapping:
image


LuizAsFight (changes requested):
- when typing incorrect network, the error icon appears (it should not)
- improve the UX on report error screen:
  - remove “Ignore” button, instead put a top-right X button to close the popup
  - remove the arrows and the X (dismiss single error), instead of having 3 horizontal buttons, we can have only 2 in horizontal way ( Dismiss, Report )

LuizAsFight (approved):

PR #1427, opened 2024-08-15
feat: update Wallet to devnet

LuizAsFight:
- Update to `fuel-core@0.33.0` and `forc@0.63.1` and `fuels@0.94.0`
  - Fixed/improved infra to run new fuel-core
    - build based on config from https://github.com/FuelLabs/chain-configuration.git
    - include `fuel_toolchain` and use `forc` instead of `fuels` utilities like `fuels-forc`
  - Use `wallet.getTransactionCost` instead of `provider.getTransactionCost`
  - Add 10% buffer on maxFee to sendTransaction, to avoid crashing on `getTxSummaryFromRequest` missing few units
  - Fixed `fuels.config.ts` to use `onDeploy` instead of `onSuccess`
  - Fixed the way to connect to contracts from `CustomAssetAbi__factory.connect(` to `new CustomAsset(`
  - Fixed manually waiting for transactions `waitForResult` after a contract call
  - Removed forced `txParams` from contract calls in e2e-contract-tests
- Cleaning on jest tests, removed some that were not worth maintaining
- Cleaning on some storybooks that were not worth maintaining
- Added `web_accessible_resources` to manifest due to error in extension setup for playwright test
- Transpile connectors in docs infra `next.config.mjs`


packages/docs/next.config.mjs
```diff
@@ -35,13 +35,15 @@ const nextConfig = {
...
     externalDir: true,
+    esmExternals: 'loose',
   },
   images: {
     dangerouslyAllowSVG: true,
     contentDispositionType: 'attachment',
     contentSecurityPolicy: "default-src 'self'; script-src 'none'; sandbox;",
   },
   trailingSlash: true,
+  transpilePackages: ['@fuels/connectors', '@fuels/react'],
```
helciofranco, line 46: do we need it? these packages are not internals

helciofranco (approved):
LGTM

packages/docs/next.config.mjs
LuizAsFight: yes we need it due to build problems

Issue #1436, opened 2024-08-28
Insert default networks no matter if they are available or not


LuizAsFight:
recently [we changed](https://github.com/FuelLabs/fuels-wallet/pull/1429) the addDefaultNetwork method to include devnet, the problem is that sometimes the devnet may be off, and this is breaking the signup flow, not allowing the user to finish the setup of his wallet after setting the password.

this issue aims to fix this problem and make the networks added directly no matter if they are online or not.

PR #1437, opened 2024-08-28
fix: add default networks regardless if they're down

helciofranco:
- Closes https://github.com/FuelLabs/fuels-wallet/issues/1436

---

| 📷 Demo |
| --- |
| <video src="https://github.com/user-attachments/assets/3d67e709-1ac4-44ca-9f44-feec93e9f041" /> |

### DIFF
--- a/packages/app/src/systems/Network/machines/networksMachine.ts
+++ b/packages/app/src/systems/Network/machines/networksMachine.ts
@@ -211,6 +211,7 @@ export const networksMachine = createMachine(
             if (!input?.data) {
               throw new Error('Invalid network input');
             }
+            await NetworkService.validateAddNetwork(input);
             const createdNetwork = await NetworkService.addNetwork(input);
             if (!createdNetwork) {
               throw new Error('Failed to add network');
--- a/packages/app/src/systems/Network/services/network.ts
+++ b/packages/app/src/systems/Network/services/network.ts
@@ -58,8 +58,6 @@ export class NetworkService {
   }
 
   static async addNetwork(input: NetworkInputs['addNetwork']) {
-    // biome-ignore lint/complexity/noThisInStatic: <explanation>
-    await this.validateAddNetwork(input);
     return db.transaction('rw', db.networks, async () => {
       const count = await db.networks.count();
       const inputToAdd = {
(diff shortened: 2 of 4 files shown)


packages/app/src/systems/Network/services/network.ts
```diff
@@ -58,8 +58,6 @@ export class NetworkService {
   }
 
   static async addNetwork(input: NetworkInputs['addNetwork']) {
-    // biome-ignore lint/complexity/noThisInStatic: <explanation>
```
arthurgeron, line 61: Removing this would stop checking if the added network is valid and compatible with the MIN_NODE_VERSION required by the wallet. Wouldn't it make more sense to create a second paremeter (e.g. `skipValidation`) so that only `addDefaultNetworks` works regardless of validation?

To make more sense with the name we could merge `addNetwork` and `validateAddNetwork` into one method, and make validation default true but optional, disabled only for the defaultNetworks call.

packages/app/src/systems/Network/machines/networksMachine.ts
```diff
@@ -211,6 +211,7 @@ export const networksMachine = createMachine(
             if (!input?.data) {
               throw new Error('Invalid network input');
             }
+            await NetworkService.validateAddNetwork(input);
```
arthurgeron, line 214: I think it'd make sense for validateAddNetwork to also select/return the network just like `addNetwork` does, so that it becomes less redundant and avoids having to use both side by side, specially because `addNetwork` is already called inside `validateAddNetwork`

arthurgeron (changes requested):

packages/app/src/systems/Network/services/network.ts
helciofranco: It's better to keep this method with a single responsibility, isn't it?
We're still validating the networks, but now the methods are called when needed.

For example, when we're adding the default networks, I'm not validating since it's from our end.
But in the other flows (which are only 2), I've added the validation accordingly.

Take a look here

https://github.com/FuelLabs/fuels-wallet/blob/26e4e14ae7f09ac2147627ed90a8a67d88520717/packages/app/src/systems/DApp/machines/addNetworkRequestMachine.tsx#L123

packages/app/src/systems/Network/machines/networksMachine.ts
helciofranco: I don't think worth for this scope changing the `validateAddNetwork` behavior. 
We need to have them working separately because we are not running the validation everywhere.

Only 3 places call this method.

2 accessible by the user. 1 called by us.

packages/app/src/systems/Network/services/network.ts
helciofranco: That way we keep the validation logic with a single responsibility, as well we keep the reusability of the original `addNetwork`.

arthurgeron (approved):

LuizAsFight (approved):

PR #1528, opened 2024-10-02
feat: check for latest url in releases and activate it

arthurgeron:
- Created ChangeNetworkService
- Checks release tags for a URL, if found replaces any existing entry or creates one, making it the selected network.
- Run checks on an async interval while the wallet is locked

### DIFF
--- a/.changeset/orange-waves-marry.md
+++ b/.changeset/orange-waves-marry.md
@@ -1,0 +1,5 @@
+---
+"fuels-wallet": patch
+---
+
+Checks release tags for mainnet and activates it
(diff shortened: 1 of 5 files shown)


LuizAsFight (changes requested):
the requirements for this is:
- it should use same logic of auto update
- it will try to manage the network every 1 minute
- it will insert network and select it in database (only if it doesn't exist yet.)
- the prop in json can be just `networkUrl` , it doesnt need to be `networks: [ ...`

LuizAsFight:
this can be closed in favor of https://github.com/FuelLabs/fuels-wallet/pull/1531

Issue #1553, opened 2024-10-06
wallet transaction list needs pagination


LuizAsFight:
we need to reduce the tx limit to 20, then lazy load the transactions after creating a pagination similar to what we have in the bridge repo transactions list

we can paginate until 200 transactions and if still has more pages then we show a button "Show all in Explorer", linking the user to the block explorer account page

PR #1558, opened 2024-10-07
feat: add tx history pagination

helciofranco:
- Closes https://github.com/FuelLabs/fuels-wallet/issues/1553

---

| 📷  Demo |
| --- |
| <video src="https://github.com/user-attachments/assets/03ce5f7e-1719-40c9-a8b3-f41317a48c00" /> |


packages/app/src/systems/Core/utils/databaseVersioning.ts
```diff
@@ -124,4 +123,30 @@ export const applyDbVersioning = (db: Dexie) => {
...
+  db.version(24).stores({
+    vaults: 'key',
+    accounts: '&address, &name',
+    networks: '&id, &url, &name, chainId',
+    connections: 'origin',
+    transactionsCursors: '++id, address, providerUrl, endCursor',
+    assets: '&name, &symbol',
+    abis: '&contractId',
+    errors: '&id',
+  });
```
arthurgeron, line 151: Can't we drop transactions table in one go?

packages/app/src/systems/Transaction/services/transaction.tsx
```diff
@@ -318,29 +315,58 @@ export class TxService {
...
+      if (hasNextPage && endCursor) {
+        cursors.push(endCursor);
+      }
+    }
+
+    return {
+      cursors,
+    };
+  }
+
```
arthurgeron, line 369: Shouldn't we limit this cache to only 200 txs?

arthurgeron (changes requested):

packages/app/src/systems/Transaction/services/transaction.tsx
helciofranco: actually I don't think it's really possible to limit exactly in `200` due to the cursor pagination.
since to get the latest page, we will need to get also the first. it means iterating all of them anyway.

i think we're fine keeping this (bridge is doing the same also)

packages/app/src/systems/Core/utils/databaseVersioning.ts
helciofranco: oh yeah, let me merge this.

packages/app/src/systems/Transaction/services/transaction.tsx
LuizAsFight: no @arthurgeron the final decision will be to get all

packages/app/src/systems/Core/utils/databaseVersioning.ts
LuizAsFight: do we need to change from `transactions` to `transactionsCursor` ? we can do it using same approach used before for changing pkey of another table in a previous upgrade. Using `transactionsTemp` then mnoving back to `transactions`

helciofranco: @LuizAsFight i did that just because of naming 
I’m storing only the cursors, not the transaction itself

helciofranco: @arthurgeron @LuizAsFight 
https://github.com/FuelLabs/fuels-wallet/pull/1558/commits/77793398cff376ac9b7898d5f44aa363c89a1100

luizstacio:
Can you fix this also? Looks really bad...
Screenshot 2024-10-08 at 01 42 21

Otherwise nice solution working really well, the first loading takes sometime but after that one, is really fast.

LuizAsFight (approved):

Issue #1263, opened 2024-04-30
Account name case sensitivity


HoTandy:
### What version of Wallet are you using?

dev Version: 0.16.2

### What is happening?

- I have 2 accounts in my wallet, I go to change the name of one of the accounts, I want to make the first letter of the word upper case rather than lower case.
- If I delete the letter `w` and replace it with `w`, there's an error message stating the name is already in use.
- If I delete the letter `w` and replace it with `W`, there's no error message and the wallet allows me to click Confirm
- Clicking `Confirm` produces an error message stating `Account Name Already Exists`
- Clicking `Confirm` again shows no error message, as if the first click crashed the wallet (but I don't think it did)
- See video below to support the explanation above

https://github.com/FuelLabs/fuels-wallet/assets/114662397/9e11b868-1f20-47e6-a9f0-0952d4796bdf

### steps to reproduce

See above and video attached

### How it should behave?

- The backend of the wallet name change functionality seems to be case insensitive as I cannot simply change a letter from lower case to upper case, it thinks the name is the same word
- The frontend however spots it's a different name and doesn't show this on the wallet screen when I change a letter in the name to upper case rather than lower case

- The frontend and back end should be consistent
- Decide whether the naming conventions should be case sensitive or not, I believe it should be 

### What os are you seeing the problem on?

Mac/Apple Desktop

### What browser?

Chrome

### What is your browser version?

Version: 0.16.2

BlackEyesForThree:
Additional explanation: The duplicate wallet name was modified again after encountering error messages, but no error messages were found

PR #1570, opened 2024-10-09
fix: account name checks FE-724

nelitow:
- Closes https://github.com/FuelLabs/fuels-wallet/issues/1263

Fixes:
- User not being able to change the account name to case-sensitive variations of the same word.
  - Example: Changing a case typo, like **Fuel wAllet** -> **Fuel Wallet** was not possible
- User not being able to change or create accounts that are substring of eachother.
  - Example: Changing **Fuel Wallet** -> **Wallet** was not possible

Considerations:
- The `filterByName` function sound like it could be used for a search filter, in which it should be case insensitive, but that is not the case at the moment, it is only used for account name editing and creation.
- Users can now create multiple accounts with similar names, which can cause confusion, changing only a letter case, so:
  - **Account 1** and **account 1** can now coexist as account names

UPDATE https://github.com/FuelLabs/fuels-wallet/pull/1570/commits/b611c8b9d1c08996bde6eb0ac7af21f9cdcbae30
Addressing an issue that arose as the function was used for account searching inside the wallet by another method, so now the account addition and editing don't rely on the same function. That was caught by e2e tests.


helciofranco (approved):
LGTM, your consideration about name conflicts is fine.

for reference, check rabby wallet, they allow even same exact name:
| Rabby Wallet |
| --- |
| Screenshot 2024-10-09 at 10 35 43 |

helciofranco:
it's missing to add a changeset @nelitow 

`pnpm changeset`

packages/app/src/systems/Account/services/account.ts
```diff
@@ -231,9 +231,9 @@ export class AccountService {
   }
 
   static filterByName(accounts: Account[], name = '') {
-    return accounts.filter((account) =>
-      account.name.toLowerCase().includes(name.toLowerCase())
-    );
+    return accounts.filter((account) => {
+      account.name === name;
```
helciofranco, line 235: it's missing a return clause here

```suggestion
      return account.name === name;
```

helciofranco (dismissed):

LuizAsFight (changes requested):
this broke the test that validates connected screen flow.


the comparison before was done with `includes` and now it's done with `===` what was the reason for changing this? the issue was to fix the lowerCase only

LuizAsFight (approved):

Issue #1564, opened 2024-10-08
Show balance with more precision


LuizAsFight:
Show precision of 6 digits<br>Do the same behavior that we have in the points program

[https://fuellabs.slack.com/archives/C07NS6ZEFK5/p1728382786667099?thread_ts=1728381709.747509&cid=C07NS6ZEFK5](https://fuellabs.slack.com/archives/C07NS6ZEFK5/p1728382786667099?thread_ts=1728381709.747509&cid=C07NS6ZEFK5)
[https://fuellabs.slack.com/archives/C07NS6ZEFK5/p1728383497110959?thread_ts=1728381709.747509&cid=C07NS6ZEFK5](https://fuellabs.slack.com/archives/C07NS6ZEFK5/p1728383497110959?thread_ts=1728381709.747509&cid=C07NS6ZEFK5)

PR #1571, opened 2024-10-09
fix: balance format precision with tooltips

helciofranco:
- Closes https://github.com/FuelLabs/fuels-wallet/issues/1564
- Closes https://github.com/FuelLabs/fuels-wallet/issues/1575

---

I've also removed the `balances` from our browser storage. Let's keep it fetching directly from the node.

| 📷  Tiny | Regular | Huge |
| --- | --- | --- |
| small amounts | not so small |huge amount |


LuizAsFight (changes requested):
this PR is changing too much, the issue aims to fix the balance only, but I see the Approve Transaction screen is showing balance with <0.00001. Revert the changes in places we don't want and keep them only for showing correct balances.

packages/app/src/systems/Account/components/BalanceAssets/BalanceAssets.tsx
```diff
@@ -45,7 +45,7 @@ export const BalanceAssets = ({
           amount={balance.amount}
           onRemove={onRemove}
           onEdit={onEdit}
-          showActions
+          showActions={false}
```
arthurgeron, line 48: default is false

LuizAsFight (changes requested):
2 problems here
- its showing wrong balance for the unknown tokens, before it was with zero units and now it's using like 9 units
- tooltip is showing only 0

image

helciofranco:
@LuizAsFight nice catch, fixes has been pushed.

LuizAsFight (approved):

PR #1577, opened 2024-10-11
Update Fuel Extension Install Link

lucemans:


lucemans:
This is where this link is shown:
image

PR #1594, opened 2024-10-13
fix: lazy trigger event wallet loaded

luizstacio:


LuizAsFight (changes requested):
broken connectors page

LuizAsFight (approved):

PR #1597, opened 2024-10-13
fix: remove delay start msg

luizstacio:


LuizAsFight (changes requested):
we need to test this better

LuizAsFight (approved):

PR #1599, opened 2024-10-13
fix: out of bounds error when opening extension

arthurgeron:
makes sure modal always to top-right


LuizAsFight (changes requested):
holding this because it can take time to get new permissions, we should plan when to release this one as we expect more delay on chrome store approval

PR #1620, opened 2024-10-18
feat: add new 'Sign' playwright helper

nelitow:
Add Sign playwright helper to use on e2e tests on connectors signs.


package.json
```diff
@@ -125,5 +125,6 @@
       "rollup@>=4.0.0 <4.22.4": ">=4.22.4",
       "fuels": "0.96.1"
     }
-  }
```
arthurgeron, line 128: We're going to use ASDF for controlling pnpm versions in the near future. So for now setting the verison in  `package.json`'s  engine should suffice

packages/playwright-utils/src/playwright-utils/fuelWalletTestHelper.ts
```diff
@@ -114,6 +114,13 @@ export class FuelWalletTestHelper {
     await approveButton.click();
   }
 
+  async walletSign() {
+    const walletPage = await this.getWalletPopupPage();
+
+    const approveButton = getButtonByText(walletPage, 'Sign');
```
arthurgeron, line 120: I think it's a good idea to check for "Sign" or other sign specific messages on the popup to make sure it's a sign message approval request

arthurgeron (changes requested):

arthurgeron (approved):

PR #1656, opened 2024-11-15
feat: bako id integration

luisburigo:
- Enable to send transaction to a name

https://github.com/user-attachments/assets/66c37e4a-8f04-4ff4-8d8a-6e178966936a

- Show name of address in transaction history
image


luizstacio (changes requested):
We are going to review this to ensure a good UX for the ecosystem.

PR #1657, opened 2024-11-18
fix: hotfix for not allowing send to ANY asset address

LuizAsFight:


helciofranco (dismissed):

packages/app/src/systems/Send/hooks/useSend.tsx
```diff
@@ -140,17 +142,32 @@ const schemaFactory = (provider?: Provider) =>
               });
             }
 
+            const assets = await AssetService.getAssets();
+            const fuelAssets = await Promise.all(
```
arthurgeron, line 146: It'll already have done fetchAll even before user inputs his password, this is not needed

arthurgeron: also, "assetsDb" is algo cached, so it won't be missing on any info

LuizAsFight: this will not fetch from server, it will just run the method to get the fuel asset of current chain and do a local verification

arthurgeron (approved):

PR #1663, opened 2024-11-23
fix: Account names should use natural sort order

alfiedotwtf:
When you have accounts `Account 1` ... all the way to `Account 15`, the list ends up looking like this:

- Account 1
- Account 10
- Account 11
- ...
- Account 15
- Account 2
- Account 3

etc.

This change makes the accounts list look like this:

- Account 1
- Account 2
- ...
- Account 9
- Account 10
- Account 11
- ...


packages/app/src/systems/Account/components/AccountList/AccountList.tsx
```diff
@@ -32,6 +32,13 @@ export function AccountList({
     setShowHidden((s) => !s);
   }
 
+  const collator = new Intl.Collator(undefined, {
+    numeric: true,
+    sensitivity: "base"
+  });
+
+  accounts = accounts.sort((a, b) => collator.compare(a.name, b.name));
```
helciofranco, line 40: can we move this to `AccountService.getAccounts`? 
so we fix the issue at the best place where the incorrect order is being returned

https://github.com/FuelLabs/fuels-wallet/blob/4fe865a6bcf4d5969b15e5e95b93b16882fa0439/packages/app/src/systems/Account/services/account.ts#L66-L70

you'll need to use `toArray()` instead of `toCollection()` in order to run `.sort()` on it (and that `sortBy('name')` can be removed).

You’ll need to use `toArray()` instead of `toCollection()` in order to run `.sort()` on it, and `sortBy('name')` will be unnecessary with this approach.


I think this approach might work:
```tsx
static async getAccounts() {
  return db.transaction('r', db.accounts, async () => {
    const accounts = await db.accounts.toArray();
    return accounts.sort((a, b) => a.name.localeCompare(b.name, undefined, { numeric: true, sensitivity: 'base' }));
  });
}
```

packages/app/src/systems/Account/components/AccountList/AccountList.tsx
```diff
@@ -32,6 +32,13 @@ export function AccountList({
...
+
   return (
```
helciofranco, line 42: also, could you add a short and clear changelog for this fix?

just run `pnpm changeset` and follow the instruction (it looks like a patch fix)

helciofranco (changes requested):
Good catch! Could we apply these two suggestions?

PR #1677, opened 2024-12-02
chore(e2e): Added support for the 'chrome-dev' channel in Playwright FE-803

nelitow:
~~Add~~ Replace `chromium` with `chrome-beta` tests as a way to catch breaking changes early on.

I would have chosen the `dev` channel as it gives us 9 to 12 weeks to catch those changes, as the `beta` channel would only give 4 to 6 weeks and the `Canary` version is too unstable, but `dev` is not available on playwright at the moment.

Decided to replace to the beta version instead of running on both as to not increase the test time. If it runs on beta we can be safe that it runs on stable.


LuizAsFight (changes requested):
working on betha doesnt mean it will work on stable, we need both.

- the stable one will be Required
- the dev one will not be Required (which means will not block the PRs if it fails)

at the end of the day, the `dev` one will be just a trigger for us to recognize we need to fix something in advance

LuizAsFight (approved):

PR #1700, opened 2024-12-11
fix: wallet reset issue

LuizAsFight:
- Closes FE-1106

Implement mechanism for backup accounts and recover in case of disaster in db (all data lost)
- synchronize db with chrome storage
- implemented recoverWallet method
- in case of disaster recover it
- simplified accountMachine
- create new database and table (outside of react) to identify if only dexiedb database is breaking
- report to sentry when there's a recover happening
- remove `shouldRecoverWelcomeFromError` logic


helciofranco (dismissed):

packages/app/src/systems/Account/services/account.ts
```diff
@@ -206,6 +207,120 @@ export class AccountService {
     });
   }
 
+  static async fetchRecoveryState() {
+    // convert to await promise.all
```
helciofranco, line 211:
```suggestion
```

packages/app/src/systems/Core/utils/database.ts
```diff
@@ -88,6 +154,9 @@ export class FuelDB extends Dexie {
 
   async clear() {
     await Promise.all([
+      chromeStorage.accounts.clear(),
+      chromeStorage.networks.clear(),
+      chromeStorage.vaults.clear(),
```
helciofranco, line 159: I've added this yesterday, so we can revert this part now

```suggestion
```

helciofranco (approved):

PR #1717, opened 2024-12-14
fix: auto-update mechanism

estherbreath:
checked for possible bug that for wallet not automatically updating and i Implement a more robust auto-update system that checks for updates regularly and notifies users when an update is available.


LuizAsFight (changes requested):
hi, this change doesn't make any sense. the PRs mentions improvement in the auto-update mechanism but the code is actually only showing a toast

PR #1718, opened 2024-12-14
test(e2e): add tests for asset visibility and management on Testnet FE-1151

nelitow:
Move the assets test to the same structure as the app tests for a cleaner setup.


arthurgeron (approved):

LuizAsFight (dismissed):
rename the files that have the `.testnet` we can remove it.

the only file that needs the `.testnet` is the yaml if it is specific for it, like we have for `tests-testnet.yaml` and `tests-devnet.yaml`. all the rest should be flexible and dont be direct for one network

rodrigobranas (approved):

arthurgeron (approved):

PR #1733, opened 2024-12-20
feat: asset value display

### LINKED ISSUE #1623: Add dollar value for assets in balance and when sending them in transaction screen
Right now the wallet doesn't show any USD value of the assets the user has. This issue aims to support it.

At the very least we need to have USD values for verified asset list (or the top coins on that list e.g. Ether, USDC, USDT, sDAI, ezETH etc.).

Should apply it on:

Send Screen

Any transaction screen (Approve, View, etc..) (they are all the same component)

* Home page balances

ShantelPeters:
closes #1623


LuizAsFight (changes requested):
can you post a screenshot in PR description of how its gonna look?

Issue #1647, opened 2024-11-06
Add integration with ENS on wallet


LuizAsFight:

khayss:
I'd love to give this a go.

I am a fullstack developer with hands-on experience building responsive and dynamic web interfaces using ReactJS, NextJS, and TypeScript. I specialize in integrating APIs and optimizing user experiences with frameworks like TailwindCSS.

mimisavage:
Is it okay if I tackle this?

chigozzdevv:
Hello, please can I jump on this task?

JoE11-y:
Could I take on this issue?

DanielEmmanuel1:
Can I work on this, please
I am a fullstack developer with great experience building responsive and dynamic UI using ReactJS, NextJS, and TypeScript. 

I’ll tackle this issue firstly by integrating ENS resolution to fetch and display the ENS name and associated avatar when available. 
For withdrawals, the wallet will replace the ETH address with the ENS name and display the ENS avatar instead of the default image, ensuring a more user-friendly and recognizable experience

0xdevcollins:
Mind if I try this one?

I am a Full Stack Blockchain Developer with proficiency in Solidity/Rust, Next.js, TypeScript, React, and Node.js. I've demonstrated my ability to adjust to a variety of requirements, perform well under pressure, and reliably produce user-centric blockchain solutions with 46 significant contributions spread over 12 OnlyDust projects.

akintewe:
I am a Full Stack Developer with a strong background in  Next.js/React, TypeScript, and Rust.  I would like to work on this issue 
Would love to newly contribute :) 
ETA is 12 hours .

Kachimercy1:
Could I be assigned to this?

1nonlypiece:
Hi ,
I’d like to work on this issue as I have experience in building frontend interfaces using frameworks like React and Next.js.


ETA:
 I estimate completing this task in about 12-24 hours. Let me know if I can get started!

ShantelPeters:
Can I take care of this issue?

Jagadeeshftw:
Hi,
I’d love to work on this issue as I have strong experience with front-end development. Please assign this to me, and I’ll deliver an efficient and secure PR within 12-24 hours.

LuizAsFight:
[...]

PR #1738, opened 2024-12-22
feat: add ENS resolution and update dependencies

0xdevcollins:
Closes: #1647


packages/app/src/systems/Account/hooks/useEns.ts
```diff
@@ -0,0 +1,72 @@
...
+        });
+        return;
+      }
+
+      setEnsData((prev) => ({ ...prev, loading: true, error: null }));
+
+      try {
+        // Using Ethereum mainnet for ENS resolution
+        const provider = new ethers.providers.JsonRpcProvider(
+          'https://mainnet.infura.io/v3/84842078b09946638c03157f83405213'
```
nelitow, line 37: Maybe something like this could be done for the URL
https://github.com/FuelLabs/fuel-explorer-private/blob/feb70258a784e390f5d61f2532666c3fc6fc5322/packages/app-commons/src/config.ts

LuizAsFight (changes requested):
looking good, can you provide in the PR description a screenshot of how its looking with the ENS and also a normal any other transaction with no ENS?

PR #1740, opened 2024-12-24
Customize lock time

OWK50GA:
Added saveLockTimeSettings function to the VaultServices and edited the autolocktimer in secret.ts

Added a component for writing and reading to/from these functions to customize the lockTime.


LuizAsFight:
in this screen we can:

1- to keep same UI pattern as other parts of the app, use the <Card>. and <CardListI.tem>. as in other parts of application
2- change the values to:
"1 minute"
"10 minutes"
"1 hour"
"1 day"
"1 week"
"never"

image

PR #1762, opened 2025-01-09
feat: add NFTs tab

### LINKED ISSUE #1686: Create NFT tab for balances in Home Screen
We currently don't have a specific NFT page / list, but only NFT badges that show up close to the asset

This issue aims to create this List of NFTs in the home page

[https://www.figma.com/design/VxEOtTLHE7hsrFsKXmNlML/Playground?node-id=3492-41391&node-type=frame&t=AJjcOvCX7X4Qgxk5-0](https://www.figma.com/design/VxEOtTLHE7hsrFsKXmNlML/Playground?node-id=3492-41391&node-type=frame&t=AJjcOvCX7X4Qgxk5-0)

helciofranco:
- Closes #1686 
- Closes `FE-1087`

---

| 📷 Demo |
| --- |
| <video src="https://github.com/user-attachments/assets/8d7a5663-9f56-4fec-942b-c02d96c47d0b" /> |


LuizAsFight (changes requested):
can we start with all collections expanded? except by the "Others" , that one can be collapsed

packages/app/src/systems/Account/components/BalanceAssets/BalanceAssets.tsx
```diff
@@ -6,36 +6,41 @@ import { AssetItem, AssetList } from '~/systems/Asset/components';
 import type { AssetListEmptyProps } from '~/systems/Asset/components/AssetList/AssetListEmpty';
 
 export type BalanceAssetListProp = {
-  balances?: CoinAsset[];
+  balances: CoinAsset[] | undefined;
```
LuizAsFight, line 9: why? can we keep with the `?` ?

packages/app/src/systems/Account/components/BalanceNFTs/groupNFTsByCollection.ts
```diff
@@ -0,0 +1,71 @@
...
+        }),
+      };
+    })
+
+    // Move "Others" to the bottom
+    .sort((a, b) => {
+      if (a.name === UNKNOWN_COLLECTION_TITLE) return 1;
+      if (b.name === UNKNOWN_COLLECTION_TITLE) return -1;
+      return 0;
+    });
```
LuizAsFight, line 68: can we sort by name also?
- A
- B
- Z
- Others

LuizAsFight (changes requested):

LuizAsFight (changes requested):
- in the black theme I can't see any separator between the collections
- I think the collection name can be a little bit bigger

packages/app/src/systems/Account/components/BalanceNFTs/groupNFTsByCollection.ts
helciofranco: Are you talking about collection names or NFT names?

If it's about **NFT names**, it's already sorting by `name`:
Screenshot 2025-01-11 at 20 18 46

If you're talking about **collection names**, I'd like to apply but I didn't because in the explorer it's also not sorting.

If you compare our listing with explorer right now, they're identical.

Thoughts?

helciofranco:
@LuizAsFight 

- in the black theme I can't see any separator between the collections
  - Fixed https://github.com/FuelLabs/fuels-wallet/pull/1762/commits/4c6ca5d22b4e5297353843dde628793dcb0cfc50
- I think the collection name can be a little bit bigger (in your video it looks bigger than when I run locally), did you change it after recording?
  - Nope. The video you've seen was the version you reviewed.. I think it's because it has a natural "scale" in the preview. I'll increase a little bit (from `$sm` to `$base`). https://github.com/FuelLabs/fuels-wallet/pull/1762/commits/76ba41bce5d4e127030103325d2247c5a39d8c87

packages/app/src/systems/Account/components/BalanceAssets/BalanceAssets.tsx
helciofranco: I don't like optionals for these cases haha

let me tell why: components like this, they will always require a `data`... 
so we will never have a `<BalanceAssets />` without setting a `balances` .
which will always force ppl to use `<BalanceAssets balances={something} />`

but i reverted here
https://github.com/FuelLabs/fuels-wallet/pull/1762/commits/2e084614e67c3d9310a46e2ae01a581f89b7aaaa

arthurgeron (changes requested):
- When I expand the top Collapsible there's some extra empty space that pushes that seems to overlap with other sections.
CleanShot 2025-01-13 at 13 49 47
CleanShot 2025-01-13 at 13 50 42


- When expanded there is also a separator that seems to be duplicated: 
CleanShot 2025-01-13 at 13 51 09

arthurgeron (approved):

LuizAsFight (changes requested):
figma: image
current: image


- smaller arrow
- slightly bigger font?
- "number" in collection should have a min-width to always be square at minimun

LuizAsFight (approved):

Issue #1719, opened 2024-12-15
ask to connect acc automatically when user change account


LuizAsFight:
if the user is connected in one account, then change to another account that is not connect, show a small popup offering to also connect that popup

OWK50GA:
The description of this issue is no clear.

It looks easy but if you're willing to exchange information on Telegram, then I'm willing to give it a go

Birdmannn:
Can I handle this. I want to improve my front end skills

OWK50GA:
May I take this issue on?

sonkeydotcom:
Can I try solving this issue?

Jagadeeshftw:
Can I start working on this?

emarc99:
Can I tackle this one?

FrankiePower:
I am Super Franky, a software developer with 3+ years of experience, and I have worked with frontend technologies like React, NextJS, JavaScript/TypeScript. I am also proficient in Cairo for smart contract development. I'm a recent graduate of Web3bridge Africa, the foremost and premier blockchain developer community. I have made contributions to many open-source projects, and I look forward to contributing to this one. My approach to this will be to:

1. Review the current account connection logic in the frontend application to understand how accounts are handled and switched.
   
2. Implement the functionality to detect when a user is connected to one account, and if they switch to another account that is not connected, trigger a small popup offering the option to connect that account.

3. Create a small, non-intrusive popup that will appear when the user switches accounts, prompting them to connect the new account to the application.

4. Test the interaction thoroughly to ensure the logic works across different browsers and platforms.

ETA: 2 days

PR #1764, opened 2025-01-10
feat: add a quick way to authorize my current account

helciofranco:
- Closes https://github.com/FuelLabs/fuels-wallet/issues/1719
- Closes `FE-1174`

---

| 📷 Demo |
| --- |
| <video src="https://github.com/user-attachments/assets/31acded5-4781-43e9-bab1-e6e8f636d3bb" /> |


LuizAsFight (commented):
this is an interesting idea but I think we would need design first to go for it, the topbar is looking a little bit crowded.

the idea of this issue is not exactly what is developed. this issue aims simply showing a toast or popup with a "connect" button so the user can connect his account from the wallet easily

helciofranco:
@LuizAsFight reverting to draft. holding until we have a design.

LuizAsFight:
@helciofranco can we complete this PR fast using a simple popup like this? Was the initial intention 
image

packages/app/src/systems/DApp/hooks/useOrigin.ts
```diff
@@ -0,0 +1,35 @@
...
+export const useOrigin = ({ url }: UseOriginProps) => {
+  const [origin, setOrigin] = useState<Origin | undefined>();
+
+  useEffect(() => {
+    if (!url) return;
+    const origin = parseUrl(url);
+    setOrigin(origin);
+  }, [url]);
+
+  return origin;
```
arthurgeron, line 34: This could be simplified into a single useMemo
```suggestion
  return useMemo(() => url ? parseUrl(url) : undefined, [url]);
```

arthurgeron (approved):
There's only one detail here

packages/app/src/systems/DApp/hooks/useOrigin.ts
helciofranco: nice catch, I missed this one thanks

LuizAsFight (approved):
when the account has assets, the connect dialog is overflowed

image

PR #1765, opened 2025-01-10
test: Improve contract e2e flakyness FE-1232

nelitow:
The goal of FE-1232 was to understand why Chrome Beta Contract tests were failing.
Turns out there were a lot of issues with the contract tests, like action timing out and the reports saying that the tests passed.
The two other main issues were the fuelWalletTestHelper failing to find the Popup page, and the aria label checks straight up not working.

While working on figuring out why the tests were failing and if they were a Chrome Beta breaking change, I took the opportunity to improve some things.

Changes:
- Make **Contract Tests** shard for improved speed
- await for `checkAriaLabelsContainsText` to avoid race condition issues
- Import `expect` in `elements.ts`.
- `checkAriaLabelsContainsText` errors fixed as we now wait for the button to trigger the Popup
- Add pollling to `waitSuccessTransaction`
- Improve logs
- `getWalletPopupPage` increased timeout and clearer logs.


LuizAsFight (commented):
checkAriaLabelsContainsText is disabled we should keep it enabled

LuizAsFight (changes requested):
hold

LuizAsFight (approved):

PR #1775, opened 2025-01-17
fix: All fixes and changes made

DanielEmmanuel1:
398944669-2d7caae6-eedc-4c92-985a-0883597cd4ae
398944665-ab362a4f-a1e4-486d-b1fa-65eb0fbaf0bb

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -15,4 +15,7 @@
   "typescript.suggest.paths": true,
   "typescript.suggest.enabled": true,
   "typescript.suggest.completeFunctionCalls": true,
+  "[typescriptreact]": {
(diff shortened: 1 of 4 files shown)


helciofranco (commented):
Thanks for your contribution.

I'm wondering if we could get the `tx mempool count` for this to decide network conditions.

Example cURL just for reference (needs to use services and xstates to follow project's pattern)
```
curl --location 'https://testnet.fuel.network/v1/graphql' --header 'Content-Type: application/json' --data '{"query":"query getTxMempoolCount {\n    nodeInfo {\n        txPoolStats {\n            txCount\n        }\n    }\n}","variables":{}}'
```

Also missing to add a changeset (`pnpm changeset`)

PR #1784, opened 2025-01-20
fix: Receive page should use Checksum

### LINKED ISSUE #1780: Receive page should use checksum
Image

Dhanraj30:
- Closes #1780 
- Closes https://linear.app/fuel-network/issue/FE-1329/receive-page-should-use-checksum

Fix receive page now use checksum address to display

here is img
Screenshot (208)


LuizAsFight (changes requested):
hi @Dhanraj30 thanks for contributing, could you create a changeset for this PR ? 
```
pnpm changeset
```

the text can be something like: `chore: make address checksummed on Receive Page`

LuizAsFight (approved):

PR #1788, opened 2025-01-21
feat: show token value converted to USD

### LINKED ISSUE #1623: Add dollar value for assets in balance and when sending them in transaction screen

### LINKED ISSUE #1792: Shown unknown assets clipped out

0.47.1

### What is happening?

"Shown unknown assets" is clipped out of bounds when user has 6+ assets

### steps to reproduce

- Have unknown assets sent to your fuel wallet account
- Have 6 or more assets

Image

### How it should behave?

It shouldn't be clipped out

Mac/Apple Desktop

### What browser?

Chrome

_No response_

arthurgeron:
- Closes #1623
- Closes #1792 
- Closes FE-1380
- Closes FE-810

- Created the convertAsset utility, which returns the token amount converted to USD.
- Token now also display their values in USD throughout Fuel Wallet.
- Moved asset endpoint data to global constants.

| Home |  |  |
|----------------------------------------------------------------------------------------------------------------------------|----------------------------------------------------------------------------------------------------------------------------|----------------------------------------------------------------------------------------------------------------------------|
 |                                                                                                                            |                                                                                                                             |


LuizAsFight (changes requested):
hold until we have design

packages/app/src/systems/Asset/services/convert-asset.ts
```diff
@@ -0,0 +1,37 @@
...
+      }),
+    });
+    const jsonResponse = (await response.json()) as ConvertAssetResponse;
+    // Fix a conversion error from the backend
+    if (jsonResponse.amount === '$0') return { amount: '$0.00' };
+    return jsonResponse;
+  } catch (_) {
+    return { amount: '$0.00' };
+  }
+}
```
LuizAsFight, line 37: do we need this call if the rate already comes in the asset and we can calculate on front end?

LuizAsFight (changes requested):

packages/app/src/systems/Asset/services/convert-asset.ts
arthurgeron: We're avoiding making the conversion in the frontend and trying to centralize that in the backend via the endpoint, both here and on explorer. 
cc: @rodrigobranas

packages/app/src/systems/Account/hooks/useAccountTotalBalance.ts
```diff
@@ -0,0 +1,15 @@
...
+    return state.context.account?.totalBalanceInUsd;
+  },
+};
+
+/**
+ * @returns The current account's total balance in USD
+ */
+export function useAccountTotalBalance() {
+  return store.useSelector(Services.accounts, selectors.totalBalanceUsd);
+}
```
LuizAsFight, line 15: why needs this if you could use

```
const { account } = useAccounts();
account.totalBalanceInUsd;
```

packages/app/src/systems/Account/components/BalanceWidget/BalanceWidget.tsx
```diff
@@ -55,13 +60,15 @@ export function BalanceWidget({
...
   const { handlers } = useAccounts();
-
+  const totalBalanceInUsd = useAccountTotalBalance() ?? 0;
   const { original, tooltip } = useMemo(() => {
     return formatBalance(account?.balance, decimals);
   }, [account]);
 
   if (isLoading || !account) return <BalanceWidget.Loader />;
 
+  const totalValue = `$${totalBalanceInUsd?.toFixed?.(2) || '0.00'}`;
```
LuizAsFight, line 70: will this have the same format behavior as before? by the looks of it it's not gonna have thousand separators.

LuizAsFight (changes requested):

packages/app/src/systems/Account/services/account.ts
```diff
@@ -107,22 +108,45 @@ export class AccountService {
...
+        ).then((rate) => {
+          assetsAmountsInUsd[asset.assetId] = rate?.amount;
+          if (rate?.amount) {
+            totalBalanceInUsd += Number.parseFloat(
+              rate.amount.replace(/[$,]/g, '')
+            );
+          }
+        });
+      });
+      await Promise.all([...assetsAmountsUsdPromise, balanceAssets]);
```
LuizAsFight, line 134: 1- we can use the rate we already have in the assets
2- if we would use this code it would be better to do the .then part later after the Promise.all completed.

LuizAsFight (changes requested):

packages/app/src/systems/Asset/components/AssetItem/AssetItemAmount.tsx
```diff
@@ -37,16 +39,30 @@ export const AssetItemAmount = ({
...
+            <Text as="span" css={styles.symbol}>
+              {symbol}
+            </Text>
+          </Text>
+          <Text
+            aria-hidden={visibility}
+            aria-label={`${symbol} conversion rate to USD`}
+            css={styles.amountInUsd}
+          >
+            {visibility ? (amountInUsd ?? '$0.00') : '•••••'}
```
LuizAsFight, line 63: like before we had "ETH ...." can we keep the $ when it's not visible also? "$ ...."

image

packages/app/src/systems/Asset/services/convert-asset.ts
LuizAsFight: we should do it in FE instead of doing many requests for a information we already have. just needs to use `bn` correctly

packages/app/src/systems/Transaction/components/TxContent/TxContent.tsx
```diff
@@ -156,7 +158,10 @@ function TxContentInfo({
         isLoading={isLoading}
       />
       {isLoading && !showDetails && <TxFee.Loader />}
-      {showDetails && !fees && <TxFee fee={tx?.fee} />}
+      {showDetails && !fees && (
+        // biome-ignore lint/suspicious/noExplicitAny: <explanation>
+        <TxFee fee={tx?.fee} tipInUsd={(tx as any)?.feeInUsd} />
```
LuizAsFight, line 163: instead of passing this "inUsd" all the time, can we just have a function that will convert it everywhere?

like 
```
convertToUsd(amount, rate, decimals)
```

LuizAsFight (changes requested):

packages/app/src/systems/Asset/components/AssetItem/AssetItemAmount.tsx
arthurgeron: Ofc!

LuizAsFight:
- "hide unknown" button is looking more gray than before. it's hard to see it. changes to this button were not needed and orthogonal to the objective of this task
- the button is not hiding when it has no unknown assets, like it was before

image

LuizAsFight:
tx fee now showing more decimal values, we need to use same logic as explorer

image
image

LuizAsFight:
this balance tooltip is showing in the incorrect place

image

PR #1800, opened 2025-01-29
fix: vite vulnerability failing audit

arthurgeron:


package.json
```diff
@@ -98,6 +98,9 @@
       "allowAny": ["react", "react-dom"],
       "ignoreMissing": ["react", "react-dom"]
     },
+    "auditConfig": {
+      "ignoreGhsas": ["GHSA-vg6x-rcgg-rjx6"]
+    },
```
LuizAsFight, line 103: do we really need this?

LuizAsFight (changes requested):

PR #1910, opened 2025-05-07
feat: return txResponse serialized when tx came from dapp

LuizAsFight:
# Checklist

- [ ] I've added error handling for all actions/requests, and verified how this error will show on UI. (or there was no error handling)
- [ ] I've reviewed all the copy changed/added in this PR, using AI if needed. (or there was no copy changes)
- [ ] I've included the reference to the issues being closed from Github and/or Linear (or there was no issues)
- [ ] I've changed the Docs to reflect my changes (or it was not needed)
- [ ] I've put docs links where it may be helpful (or it was not needed)
- [ ] I checked the resulting UI both in Light and Dark mode (or no UI changes were made)
- [ ] I **reviewed** the **entire PR** myself (preferably, on GH UI)

### DIFF
--- a/.changeset/gorgeous-seahorses-bathe.md
+++ b/.changeset/gorgeous-seahorses-bathe.md
@@ -1,0 +1,5 @@
+---
+"fuels-wallet": patch
+---
+
+feat: return txResponse serialized when tx came from dapp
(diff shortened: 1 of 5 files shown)


LuizAsFight (commented):
blocking this PR for now as it can be a breaking change for applications using old connectors versions. reasoning is that we'll return a TransactionResponse while the connector is waiting for a string.

one idea to mitigate this problem is to receive the connector version(or a flag prop asking for the txResponse) in the wallet and then decide if returns the string as of the old versions, or return the serialized TransactionResponseJson

LuizAsFight:
closing in favor of https://github.com/FuelLabs/fuels-wallet/pull/1920

PR #1928, opened 2025-05-22
fix: create migration to convert old addresses with bech32

### LINKED ISSUE #1801: Bech32 address migration didn't work for everyone
Some users have been reporting that they are still seeing `fuel...` addresses.
This happened to me last week when I used a very old Chrome profile in my computer (so my wallet for some reason also didnt' run the address migration).

| 📷 Example |
| --- |
| Image |

### Idea to reproduce:
1. Install a old wallet version (with `bech32` addresses)
2. Check if the address migration will run

LuizAsFight:
- Closes #1801 
- Closes FE-1413

# Checklist

- [x] I've added error handling for all actions/requests, and verified how this error will show on UI. (or there was no error handling)
- [x] I've reviewed all the copy changed/added in this PR, using AI if needed. (or there was no copy changes)
- [x] I've included the reference to the issues being closed from Github and/or Linear (or there was no issues)
- [x] I've changed the Docs to reflect my changes (or it was not needed)
- [x] I've put docs links where it may be helpful (or it was not needed)
- [x] I checked the resulting UI both in Light and Dark mode (or no UI changes were made)
- [x] I **reviewed** the **entire PR** myself (preferably, on GH UI)


helciofranco (dismissed):

nelitow (commented):
The problem still happens in my tests.
image
Reproduction:
Installed wallet 0.21.3, then upgraded to this branch

LuizAsFight:
> The problem still happens in my tests. image Reproduction: Installed wallet 0.21.3, then upgraded to this branch

This scenario is not likely to happen, because the user would need to be very far in an old version. this is happening because the old migration uses `fromDynamicInput` , I can change that one to use the new method as well. 


nice catch

nelitow (approved):
Performed the same test again and now it worked as expected.

